PR #8, opened 2017-02-01
Add installation docs

### LINKED ISSUE #5: Installation Guide
Hey there,

The README is missing a simple installation guide. I'm very interested in using this in the future, but I don't want people to get their hopes up when using something like Exponent thinking that they can use it like any JS library (before they look at the code and see the *.gradle files of course). I'm assuming it's just a simple `npm i -S lottie-react-native && react-native link lottie-react-native`, but there are often caveats to native modules that could be explained if they exist with this one.

Thanks and great job guys!

### LINKED ISSUE #7: How to run the example?
Not really know how to run the example. If have a more explicit instruction is better 😬

lelandrichardson:
Fixes #5 
Fixes #7

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,29 @@ respective libraries:
 [Lottie for Android](https://github.com/airbnb/lottie-android)
 
 
+## Installation
+
+Get started with Lottie by installing the node module:
+
+```bash
+npm i --save lottie-react-native
+```
+
+From here, RN's built in linking should work:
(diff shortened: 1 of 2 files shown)


AlexDM0:
This does not add the actual libs if you're not using pods. See #19 and #23 for issues popping up by following the installation guide. Maybe add Carthage support for the lottie-ios?

PR #15, opened 2017-02-02
refactor: remove unused props

astroplex:


ljharb:
Although React makes this not necessary, it's quite idiomatic to keep it. Is there any benefit to removing it?

astroplex:
There seems to be a difference in coding style. You can close it if you do not need it. :)

ljharb:
We have internal linting rules that require that React component constructor arguments be present. Thanks for the contribution!

PR #20, opened 2017-02-02
[iOS] Fix header search paths 

### LINKED ISSUE #12: iOS Headers not compatible with RN 0.40+
Tried installing in my existing (and also new) project and I'm getting
```
'RCTView.h' not found
'RCTViewManager.h' not found
```

jhen0409:
`../../node_modules` is only work for example.

Also, I can't found `Lottie.framework` if I just use `react-native link` without CocoaPods, so it still not work, could we have a download link for that?


jhen0409:
Looks like #12 is pointing to RN 0.40 problem, so I removed it on top. This is fix for RN <= 0.39.

lelandrichardson:
Thanks for this. I'm looking into ios header issues and how it related to the 0.40 stuff. I might have a separate PR that fixes both of these problems, but if not, I'll pull this in.

PR #22, opened 2017-02-02
updated podspec summary

AlexDM0:
I think this line was left over from the initial copy paste session :)


lelandrichardson:
:man_facepalming:

PR #36, opened 2017-02-04
Change classes prefix to LRN instead of Lottie on iOS

janicduplessis:
3 letters prefixes are the standard for iOS classes, LRN for Lottie React Native.

This doesn't change the native module name so no breaking change.


lelandrichardson:
I will push out a release later today w/ the updates

Issue #32, opened 2017-02-03
duplicate interface definition for class 'RCTView'


zawarudo:
`
In file included from ./node_modules/lottie-react-native/lib/ios/LottieReactNative/LottieContainerView.m:9:
In file included from ./node_modules/lottie-react-native/lib/ios/LottieReactNative/LottieContainerView.h:12:
../../../react-native/React/Views/RCTView.h:21:1: error: duplicate interface definition for class 'RCTView'
`

lottie-react-native: 1.0.1,
react-native 0.41.0,

Full error log: https://gist.github.com/zawarudo/ec17df88555b58a483ed81c86e98e171

Similar to
https://github.com/airbnb/react-native-maps/issues/937
https://github.com/airbnb/react-native-maps/issues/957

ferologics:
fixed in #28

kyle-ssg:
Unless I'm missing something, the above PR didn't solve the problem. I'm still getting the issue mentioned on 0.40.0+ projects

lelandrichardson:
You're getting this error because CocoaPods is referencing react native from the cocoapods specs repository, and not your local node_modules folder.  I'm assuming this is happening because you're including react into your project as a static library (not with cocoapods), but including Lottie with cocoapods.

I'm hoping to set this project up a little bit differently so it is easier to include as a static library ASAP. At the moment, though, you can add LottieReactNative.xcodeproject as a static library, and then also add Lottie as a static library, or you can change your project to install react via cocoapods.

Apologies for the back and forth on this

zawarudo:
@lelandrichardson 

Thanks for the clarification, I'm still having trouble understanding what you mean however.

>or you can change your project to install react via cocoapods.
Should I ommit the react entry to allow CocoaPods to handle the installation?
Below is my PodFile and output when installing.


**Podfile**
> 
> target "Example" do
> 
>   pod 'React', :path => '../node_modules/react-native'
> 
>   pod 'BVLinearGradient', :path => '../node_modules/react-native-linear-gradient'
> 
>   pod 'lottie-react-native', :path => '../node_modules/lottie-react-native'
> 
> end

**Output**
> 
> Fetching podspec for `BVLinearGradient` from `../node_modules/react-native-linear-gradient`
> Fetching podspec for `React` from `../node_modules/react-native`
> Fetching podspec for `lottie-react-native` from `../node_modules/lottie-react-native`
> Downloading dependencies
> Using BVLinearGradient (2.0.0)
[...]

PR #41, opened 2017-02-07
Fix build failures caused by RN0.4

offwhite:


zawarudo:
I was trying this myself to plow through the issues but is it really advisable to remove the import check statements?

lelandrichardson:
@offwhite this doesn't look correct. If this is causing issues for you, we need to solve it a different way. Perhaps the order needs to change?

zawarudo:
@lelandrichardson This is also what I had to do to bypass issue #32 but it gave me new issues about not being able to import lottie.h

lelandrichardson:
@zawarudo hang tight. We are working on a solution to this. Fix is incoming, but will take place in lottie-ios

Issue #59, opened 2017-02-13
Cannot build on React Native 0.41.2


yoohoogun114:
# Current behavior:
Cannot build.

# Reproduce:
$ React native awesome
$ cd awesome
$ npm i --save lottie-react-native
```bash
└─┬ lottie-react-native@1.0.6 
  ├── lottie-ios@1.0.2 
  └─┬ react-native-safe-module@1.1.0 
    └── dedent@0.6.0 
```
$ react-native link lottie-ios
```bash
rnpm-install info Linking lottie-ios ios dependency 
rnpm-install info iOS module lottie-ios has been successfully linked 
```
$ react-native link lottie-react-native/
```bash
rnpm-install info Linking lottie-react-native/ android dependency 
rnpm-install info Android module lottie-react-native/ has been successfully linked 
rnpm-install info Linking lottie-react-native/ ios dependency 
rnpm-install info iOS module lottie-react-native/ has been successfully linked 
```

$ react-native run-ios
```bash
ld: warning: object file (/Users/yuhogyun/test/awesome/ios/build/Build/Products/Debug-iphonesimulator/libLottie.a(LOTAnimatableLayer.o)) was built for newer iOS version (10.2) than being linked (8.0)
ld: warning: object file (/Users/yuhogyun/test/awesome/ios/build/Build/Products/Debug-iphonesimulator/libLottie.a(LOTEllipseShapeLayer.o)) was built for newer iOS version (10.2) than being linked (8.0)
ld: warning: object file (/Users/yuhogyun/test/awesome/ios/build/Build/Products/Debug-iphonesimulator/libLottie.a(LOTGroupLayerView.o)) was built for newer iOS version (10.2) than being linked (8.0)
ld: warning: object file (/Users/yuhogyun/test/awesome/ios/build/Build/Products/Debug-iphonesimulator/libLottie.a(LOTLayerView.o)) was built for newer iOS version (10.2) than being linked (8.0)
ld: warning: object file (/Users/yuhogyun/test/awesome/ios/build/Build/Products/Debug-iphonesimulator/libLottie.a(LOTMaskLayer.o)) was built for newer iOS version (10.2) than being linked (8.0)
ld: warning: object file (/Users/yuhogyun/test/awesome/ios/build/Build/Products/Debug-iphonesimulator/libLottie.a(LOTRectShapeLayer.o)) was built for newer iOS version (10.2) than being linked (8.0)
[...]

PR #60, opened 2017-02-14
Unfreeze pods versions in iOS Example

huangbow:
example/ios/Podfile.lock keeps the lottie-ios to v1.0.0, which leads many weird problems, and partially fix...such as #59 #33 etc.


lelandrichardson:
I'm confused. This is for the example project and shouldn't affect people pulling in this library for use in their app?

To update this file we should instead just run `pod install` and the proper lottie-ios version will get updated?

lelandrichardson:
Closing in favor of #97 and #98

PR #68, opened 2017-02-17
Adds framework search paths for Lottie.framework installed with carthage

arthurjamain:
Hey there, 

This PR is somewhat dirty in that I did not think about process / style etc. However it solves the following problem : 

- Install lottie-ios with Carthage
  - Make the framework available within the project witht he usual carthage manipulations
- install lottie-react-native with npm
  - link etc

Problem : the files within the lottie-react-native subproject cannot access the contents of Lottie.framework because it does not know where to find them

Solution : add this somewhat standard path to the Framework Search Path of the lottie-react-native subproject so that it can access and use them.

I feel like there are more fit solutions to this problem that could be expressed by actual xcode / ios experts, which I'm not. This did solve the issue though.

If you feel like this should be a thing but the PR is dirty, please advice, I will absolutely clean up :)

P.S. This is the only acceptable approach for me because my build is handled by a CI server that reinstalls every npm dependency for each build.

### DIFF
--- a/lib/ios/LottieReactNative.xcodeproj/project.pbxproj
+++ b/lib/ios/LottieReactNative.xcodeproj/project.pbxproj
@@ -226,11 +226,17 @@
 		11FA5C5B1C4A1296003AC2EE /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(SRCROOT)/../../../ios/Carthage/Build/iOS/**",
+					"$(SRCROOT)/../../../../ios/Carthage/Build/iOS",
+				);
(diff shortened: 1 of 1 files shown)


lelandrichardson:
Thanks for the PR!  I'm not quite sure what the right thing to do here is either...

hmm.  @bachand @buba447 do you have any advice?

bachand:
Hey @arthurjamain I did some thinking about this. AFAIK when you pull in `lottie-react-native` via NPM it pulls in a version of `lottie-ios` right ( @lelandrichardson keep me honest)? Would it be possible to use that version instead of importing `lottie-ios` via Carthage? Could be missing something though so if so maybe you can help me better understand your needs!

gpeal:
@arthurjamain Do you still need this?

PR #97, opened 2017-03-26
Fix example project

bachand:
The `Podfile` in the example project referenced `lottie-ios` version 1.0.0 but in 1.0.1 we changed the prefix from "LA" to "LOT."" So currently the example project doesn't build.

I've added a `Gemfile` and updated the `npm` command to run CocoaPods via `bundle exec`, which ensures the right version of CocoaPods.

There is some risk that user doesn't have Bundler but I think it should be fairly straightforward as to how to get it.

I ran the example project. It looks odd on iPad but I expect that was the case before. Here it is on iPhone
lottie-rn


lelandrichardson:
@bachand I just updated this in  #98 (which this is now conflicting with), although I didn't add a gemfile and changes to use the bundler. Think you can rebase on top of these changes?

bachand:
@lelandrichardson I rebased and then ran through the steps again and it seems to be working!

lelandrichardson:
Thanks!

PR #129, opened 2017-05-07
Reading all numbers as double from ReadableMap and ReadableArray

thotegowda:
One of our AfterEffect json file was not working properly in lottie-react-native. But was working fine in lottie-react-native. Verified it by placing json in to official examples. 

After spending some time in code, figured out the problem. The problem is `JSONReadableMap.java` and `JSONReadableArray.java` files are converting `Number` type to `Integer` by default and reading as `Double` on exception. This seems to be causing some issues and stripping off decimal values.

```java
  @Override
  public Object opt(String name) {
    switch (map.getType(name)) {
    ...
      case Number:
        try {
          return map.getInt(name);
        } catch (Exception e) {
          return map.getDouble(name);
        }
    ...
    }
  }
```

Got animation working fine by changing the default type to double. 

```java
  @Override
  public Object opt(String name) {
    switch (map.getType(name)) {
    ...
      case Number:
          return map.getDouble(name);
    ...
    }
  }
```

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/lottie/JSONReadableArray.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/lottie/JSONReadableArray.java
@@ -148,11 +148,7 @@ class JSONReadableArray extends JSONArray {
         case Boolean:
           return getBoolean(index);
         case Number:
-          try {
-            return getInt(index);
-          } catch (Exception e) {
             return getDouble(index);
-          }
         case String:
           return getString(index);
         case Null:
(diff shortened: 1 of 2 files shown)


thotegowda:
Does anyone know the timeline for someone to look at this PR and see if it makes sense to merge and release?  Reason for asking: If it doesn't fall within our release cycle, will have to implement stop gap solution until this gets pushed to release.

twitwi:
I also have animations that were failing to play and that play if I apply this PR.

sonaye:
@lelandrichardson can we have this merged? Lottie is practically unusable on Android without it.

sonaye:
@gpeal Can you help with this? it's been two months.

PR #132, opened 2017-05-12
Feature/finish callback

maxkomarychev:
Add a callback prop that is called when animation is finished.

partially relevant to https://github.com/airbnb/lottie-react-native/issues/3

### DIFF
--- a/lib/android/src/main/java/com/airbnb/android/react/lottie/ReactLottieAnimationView.java
+++ b/lib/android/src/main/java/com/airbnb/android/react/lottie/ReactLottieAnimationView.java
@@ -1,0 +1,33 @@
+package com.airbnb.android.react.lottie;
+
+import android.animation.Animator;
+import android.content.Context;
+
+import com.airbnb.lottie.LottieAnimationView;
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.uimanager.events.RCTEventEmitter;
+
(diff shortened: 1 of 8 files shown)


josephphayden:
What's the status of this? Very keen for this feature :)

lib/android/src/main/java/com/airbnb/android/react/lottie/ReactLottieAnimationView.java
```diff
@@ -0,0 +1,33 @@
...
+import android.animation.Animator;
+import android.content.Context;
+
+import com.airbnb.lottie.LottieAnimationView;
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.uimanager.events.RCTEventEmitter;
+
+public class ReactLottieAnimationView extends LottieAnimationView implements Animator.AnimatorListener {
```
varunrau: why is this class necessary? why can't `LottieAnimationView` implement `Animator.AnimatorListener`

luisfuertes:
lottie-react-native already has this feature?

gpeal:
This will need to be augmented in a few ways to handle pause, cancel, repeat, and other events as well as the merge conflicts since this. Since this is now pretty old, I'm going to close this particular PR but either you can put up a new one or I'll work on this soon.

maxkomarychev:
I ended up using simple wrapper with `Animated`-driven animations. Proposed solution is no longer relevant and in fact looks too complicated.

PR #161, opened 2017-07-27
Use prop-types package (PropTypes in React package is deprecated)

jrnk:
Because `PropTypes` is not available anymore in newer React packages, we need to use the standalone `prop-types` package to keep using PropTypes functionality.

See https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes


fermuch:
@gpeal any chance to release a new version with this fix and support for react@16?

PR #162, opened 2017-07-27
Move PropTypes from react core to prop-types library

thotegowda:


thotegowda:
There is already a PR for this

PR #168, opened 2017-08-02
Feature/support react 0.47

maxkomarychev:
As of `react-native` relase v0.47.0 `createJSModules` method is no
longer available.

Reference: https://github.com/facebook/react-native/releases/tag/v0.47.0


kelset:
Don't want to flame, but #169 is a better solution :/

maxkomarychev:
fine with me, backward compatibility is good.
I'll close mine in favor of #169

PR #174, opened 2017-08-17
Iterators supported inside of the JSONReadableMap

dryganets:
keys method of JSONReadableMap implemented.


dryganets:
We are using this method in our implementation of template based animations.
I'm going to contribute this as a set of small PRs.

Abstract:
We have a lot of different animations in our app.
To make animations perform well, we had to move all animation specs to assets.
As some of our animations differ only by a couple of fields, we added a capability to pass template context from js side. This way we could have a single file instead of multiple copies of the same asset.

PR #189, opened 2017-09-13
Expose a new prop to enable merge paths on android

tlenclos:
This should fix https://github.com/airbnb/lottie-react-native/issues/185.

I don't have the warning saying `Animation contains merge paths but they are disabled.` but sadly I tried with an animation with merge paths that is working on iOS and it only rendered a blank space on android.

### DIFF
--- a/example/LottieAnimatedExample.js
+++ b/example/LottieAnimatedExample.js
@@ -104,6 +104,7 @@ export default class LottieAnimatedExample extends React.Component {
             }}
             source={EXAMPLES[this.state.example].getJson()}
             progress={this.state.progress}
+            enableMergePaths={true}
           />
         </View>
       </View>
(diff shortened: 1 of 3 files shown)


zolrath:
Could this not be automatically enabled if `(Platform.OS === 'android' && Platform.Version === 19)`? Are there any disadvantages to enabling merge paths on android platforms that support it?

gpeal:
Yes because the appearance may be vastly different and if an engineer puts
in an animation and it looks correct to them, they may not realize that
it's totally broken for some of their users.

gpeal:
@tlenclos @zolrath

tlenclos:
Ok so `enableMergePathsAndroidForKitKatAndAbove` ? 😛

gpeal:
@tlenclos Yeah, shitty long name I know but anything less would be misleading IMO.

tlenclos:
Alright, I amended the last commit and rebased on master 😄

PR #212, opened 2017-10-17
Create typings file for lottie-react-native

dannycochran:
This maps the documentation from http://airbnb.io/lottie/react-native/props.html to TypeScript typings, with some additional information provided.

I'm not sure if the "AnimationObject" interface is entirely accurate -- for instance, all those fields are currently marked as required but perhaps they can be optional?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,64 @@
+import * as React from 'react';
+
+import {
+  Animated,
+  StyleProp,
+  ViewProperties,
+  ViewStyle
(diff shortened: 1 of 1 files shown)


dannycochran:
@lelandrichardson may I have a review here?

Also per https://github.com/airbnb/lottie-react-native/issues/217 (ignoring any of the more snide comments), what are plans for ongoing maintenance? I know it's a lot of work but this is an awesome library and I'd love to see it maintained going forward.

PR #223, opened 2017-11-09
[fix] eliminate warning on older RN version

erichua23:
In my case, I use react native 0.40.0, it shows "Warning: Failed prop type: Cannot read property 'style' of undefined".

### DIFF
--- a/lib/js/Animation.js
+++ b/lib/js/Animation.js
@@ -34,7 +34,7 @@ const ViewStyleExceptBorderPropType = (props, propName, componentName, ...rest) 
       'to render a border around this component, wrap it with a View.'
     );
   }
-  return ViewPropTypes.style(props, propName, componentName, ...rest);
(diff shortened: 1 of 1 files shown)


lib/js/Animation.js
```diff
@@ -34,7 +34,7 @@ const ViewStyleExceptBorderPropType = (props, propName, componentName, ...rest)
       'to render a border around this component, wrap it with a View.'
     );
   }
-  return ViewPropTypes.style(props, propName, componentName, ...rest);
+  return (ViewPropTypes || View.propTypes).style(props, propName, componentName, ...rest);
```
gpeal: This will cause warnings on new RN versions
https://github.com/facebook/react-native/blob/e10f7788c4ebf29d96594f9753264bdf6146f0cf/Libraries/ReactNative/verifyPropTypes.js#L41

gpeal:
@erichua23 bump

Issue #271, opened 2018-01-24
It seems the type definition for typescript is not working...


rodrigoelp:
Hi guys, I think the type definitions for typescript are wrong or something is missing...

Today I wrote this app

```javascript
/**
   * Contents of index.js
   */
import React, { Component } from 'react';
import { View, Animated, AppRegistry } from 'react-native';
import LottieView from "lottie-react-native";

class App extends Component<{}, {}> {
  constructor(props) {
    super(props);

    this.state = {
      animatedProgress: new Animated.Value(0)
    };
  }

  componentDidMount() {
    Animated.timing(this.state.animatedProgress, { toValue: 1, duration: 3000 }).start();
  }

  render() {
    return (
      <View style={{ flex: 1 }}>
        <LottieView style={{flex: 1}} source={require('./lib/animation.json')} progress={this.state.animatedProgress} />
      </View>
    );
  }
}

AppRegistry.registerComponent('testlottie', () => App);
```

Pretty straightforward, this code works without issues... Decided to write it in typescript

```typescript
/**
   * Contents of index.tsx
   */
import * as React from "react";
import { View, AppRegistry, Animated } from "react-native";
import LottieView from "lottie-react-native";

class TypedApp extends React.Component<{}, { animatedProgress: Animated.Value }>{
    constructor(props: any) {
        super(props);

        this.state = {
            animatedProgress: new Animated.Value(0)
        }
    }

    componentDidMount() {
        Animated.timing(this.state.animatedProgress, { toValue: 1, duration: 3000 })
            .start();
    }

    render() {
        return (
            <View style={{ flex: 1, backgroundColor: "black" }}>
                <LottieView source={require("./animation.json")} progress={this.state.animatedProgress} />
            </View>
        );
    }
}

AppRegistry.registerComponent('testlottie', () => TypedApp);
```

You compile this code and it generates the `index.js` file... running the application shows the following error:

![image](https://user-images.githubusercontent.com/301434/35316316-0efc8de8-0125-11e8-8ebb-32a2fca980df.png)

The reason why we get this error is that the type definition is trying to reference a constructor that does not exist... Checking the generated `index.js` I can see the error

```javascript
    render() {
        return (React.createElement(react_native_1.View, { style: { flex: 1, backgroundColor: "black" } },
            React.createElement(lottie_react_native_1.default, { source: require("./animation.json"), progress: this.state.animatedProgress })));
            // The line above should not have the ".default", it should be 
[...]

PR #273, opened 2018-01-24
Fixed type definitions for typescript and extended the examples to include a typescript project

rodrigoelp:
As indicated in the issue #271, the typescript definition file does not produce a valid output when trying to compile the code and execute it.

The root cause of this issue is the way the class is exported via `export default` in the module, creating an access to the namespace as opposed to the prototype as such. This generates an `invariant violation: element type is invalid`

Based on how this library is written, creating a type definition with a default export as indicated generates the following code when compiled
``` javascript
render() {
     return (React.createElement(react_native_1.View, { style: { flex: 1, backgroundColor: "black" } },
         React.createElement(lottie_react_native_1.default, { source: require("./animation.json"), progress: this.state.animatedProgress })));
}
```

Instead of:
``` javascript
render() {
     return (React.createElement(react_native_1.View, { style: { flex: 1, backgroundColor: "black" } },
         React.createElement(lottie_react_native_1, { source: require("./animation.json"), progress: this.state.animatedProgress })));
}
```
The difference being: `.default` after `lottie_react_native_1`

In order to fix this issue, the type definition was rewritten and an example was created to illustrate the appropriate use of the library for typescript.
To avoid misunderstanding, the example folder was replaced with an `examples` folder containing the previous javascript example and a new folder containing an almost identical copy written for typescript can be found in the `typescript` folder.


rodrigoelp:
Optionally, I could create a pull request that just contains the typescript definition update and the example can be done as part of the readme file

dannycochran:
LGTM as author of the file originally. It might make the review go faster to just update the TS definition file, though.

Pinging @gpeal -- it'd be good to get this merged and released, it blocks easy usage of lottie-rn with TypeScript.

rodrigoelp:
Is there any update on this?

gpeal:
@rodrigoelp I don't feel comfortable duplicating the example project because it'll just become two project that have to be updated for any change. I would be okay if the sample project itself was ts though so we could just convert it in place rather than duplicate it.

rodrigoelp:
Hi @gpeal, understood. Should I change the pull request? Has LGTM review the type definition? in that case I would be happy to do the changes and revert back to a single example

gpeal:
@rodrigoelp I'm happy to stamp the type definition changes on their own and you can do the sample as a separate PR.

rodrigoelp:
Ok @gpeal, I will update the pull request to just have the type definition 👍

rodrigoelp:
Closing this pull request as I have created a new one here #291

PR #291, opened 2018-03-06
Fixed typescript support

rodrigoelp:
This change has a side effect as the typescript code is not going to be identical to the javascript sample because of the way the export works in Javascript.


rodrigoelp:
Hi @gpeal and @dannycochran, here you will find just the type definition for typescript and a small change in `package.json` to allow it to recognise both the js and ts typings.

dannycochran:
@gpeal can you push a new release@2.3.3?

PR #296, opened 2018-03-15
Bump Lottie to 2.5

Minishlink:
Example app works great :)


gpeal:
Thanks @Minishlink! Why did so many npm package tarballs get added since you only changed lottie?

If you @ mention me, I'll get an email and can respond more quickly btw.

dannycochran:
@gpeal - is it possible to take master and publish a new version to npm? It doesn't seem possible to use a commit hash from this repo since I believe there's some type of compilation you do prior to publishing to npm.

PR #298, opened 2018-03-17
cacheStrategy property exposed to js layer.

dryganets:
As cacheStrategy property dependent on the animationName we need to use property manager pattern.

I decided to apply the pattern to all properties for android.
If you think this is too much I could roll back not related changes.

Ideally should wait for:
https://github.com/airbnb/lottie-android/pull/691
to be completed.

### DIFF
--- a/src/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManager.java
+++ b/src/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManager.java
@@ -3,7 +3,6 @@ package com.airbnb.android.react.lottie;
 import android.os.Handler;
 import android.os.Looper;
 import android.support.v4.view.ViewCompat;
-import android.util.Log;
 import android.widget.ImageView;
 
 import com.airbnb.lottie.LottieAnimationView;
@@ -13,16 +12,19 @@ import com.facebook.react.uimanager.SimpleViewManager;
 import com.facebook.react.uimanager.ThemedReactContext;
 import com.facebook.react.uimanager.annotations.ReactProp;
 
-import org.json.JSONObject;
-
 import java.util.Map;
+import java.util.WeakHashMap;
 
 class LottieAnimationViewManager extends SimpleViewManager<LottieAnimationView> {
-  private static final String TAG = LottieAnimationViewManager.class.getSimpleName();
+  public static final String TAG = LottieAnimationViewManager.class.getSimpleName();
+
   private static final String REACT_CLASS = "LottieAnimationView";
   private static final int VERSION = 1;
   private static final int COMMAND_PLAY = 1;
   private static final int COMMAND_RESET = 2;
+
+  // The idea here is to keep property manager for the view as long as the view is alive
+  private Map<LottieAnimationView, LottieAnimationViewPropertyManager> propManagersMap = new WeakHashMap<>();
 
   @Override public Map<String, Object> getExportedViewConstants() {
     return MapBuilder.<String, Object>builder()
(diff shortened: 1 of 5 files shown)


src/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManager.java
```diff
@@ -13,17 +12,20 @@
...
 class LottieAnimationViewManager extends SimpleViewManager<LottieAnimationView> {
-  private static final String TAG = LottieAnimationViewManager.class.getSimpleName();
+  public static final String TAG = LottieAnimationViewManager.class.getSimpleName();
+
   private static final String REACT_CLASS = "LottieAnimationView";
   private static final int VERSION = 1;
   private static final int COMMAND_PLAY = 1;
   private static final int COMMAND_RESET = 2;
 
+  // The idea here is to keep property manager for the view as long as the view is alive
```
gpeal: Don't leave comments like this in the code. Put them in the PR.

src/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.java
```diff
@@ -0,0 +1,142 @@
...
+    this.imageAssetsFolder = imageAssetsFolder;
+  }
+
+  public void setEnableMergePaths(boolean enableMergePaths) {
+    this.enableMergePaths = enableMergePaths;
+  }
+
+  /**
+   * Updates the view with changed fields.
+   * Majority of the properties here are indepenbent so they are has to be reset to null
```
gpeal: type in independent

src/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.java
```diff
@@ -0,0 +1,142 @@
...
+   * as soon as view is updated with the value.
+   *
+   * The only exception from this rule is the group of the properties for the animation.
+   * For now this is animationName and cacheStrategy. These two properties are should be set
+   * simultaneously if the dirty flag is set.
+   */
+  public void commitChanges() {
+    LottieAnimationView view = viewWeakReference.get();
+    if (view != null) {
```
gpeal: Just return early if view == null so the entire method doesn't have to me nested

src/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  public void setAnimationName(String animationName) {
+    this.animationName = animationName;
+    this.animationNameDirty = true;
+  }
+
+  public void setAnimationJson(String json) {
+    try {
+      this.animationJson = new JSONObject(json);
```
gpeal: Hold off on submitting this until #296 is merged so you can keep the animation as a string since 2.5.0 uses a JsonReader

src/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.java
```diff
@@ -0,0 +1,142 @@
...
+  public void commitChanges() {
+    LottieAnimationView view = viewWeakReference.get();
+    if (view != null) {
+      if (animationJson != null) {
+        view.setAnimation(animationJson);
+      }
+
+      if (animationNameDirty) {
+        view.setAnimation(animationName, cacheStrategy);
+        animationNameDirty = false;
```
gpeal: Why do the properties have to be reset? Shouldn't calling commit multiple times apply the same values?

dryganets: React sends only changed property values over the bridge.
That's why there is a slight difference between independent and dependent properties.

For independent - you want to reset them to don't waste native calls without need. If you leave it cached you will update it on every unrelated update which is different from default behavior.
For dependent - you will lose part of the information if you reset them, so you need to have a separate dirty flag.

You could see that I'm using boxed types for primitive values - it is to distinguish values been set from not set, kind of field's personal dirty flag.

src/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.java
dryganets: ok!

dryganets:
@gpeal the PR is updated with JsonReader and LottieView.DEFAULT_CACHE_STRATEGY.

itshnegg:
@gpeal could you please review this pull request?

itshnegg:
The code looks good. @gpeal could you please merge it?

PR #352, opened 2018-07-16
Use new dependency configurations

kevinmeresse:
'provided' and 'compile' are now obsolete and will be removed soon


emilioicai:
this was included in PR #354

PR #358, opened 2018-07-29
Autoplay animations when animation's source has changed

Minishlink:
When having a LottieView with `autoPlay`, if you changed the source, the animation was not auto played.


s349856186:
@Minishlink I used "lottie-react-native": "^2.5.11",  but not work, the animation still stop when source had changed. How can I fix it?

PR #363, opened 2018-08-08
Preserve content mode when animation view is replaced

sharplet:
Because we're setting the content mode directly on `_animationView` (see commit 89775d9f6e2b7d768771f43b62454c0d154412c3), we need to copy it when the animation view is replaced.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "types": "src/js",
   "scripts": {
     "start": "node node_modules/react-native/local-cli/cli.js start",
+    "prepare": "npm run build",
     "build": "babel src/js --out-dir lib",
     "watch": "babel src/js --out-dir lib --watch",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -6,6 +6,7 @@
   "types": "src/js",
   "scripts": {
     "start": "node node_modules/react-native/local-cli/cli.js start",
+    "prepare": "npm run build",
```
emilioicai: what's the purpose of this script?

sharplet: Adding it allowed me to install from this branch directly without a published release.

fedegl:
@emilioicai Would it be possible to get this merged in? I am right now having to use the fork 😬 and would prefer to switch back to an official release.

emilioicai:
@fedegl done :)

fedegl:
@emilioicai Awesome thanks!

PR #374, opened 2018-08-26
Add duration prop

Minishlink:
Forgot to add it from [easy-lottie-react-native](https://github.com/bamlab/easy-lottie-react-native)


emilioicai:
thanks @Minishlink !

PR #384, opened 2018-09-21
Remove Gradle warning

bahaa96:
WARNING: Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
It will be removed at the end of 2018. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html


emilioicai:
Thanks @bahaa96, this is indeed needed but I cannot compile after making these changes. Did you test the build was working fine?

PR #389, opened 2018-10-03
Fixed typescript type definition export for 3.x

rodrigoelp:
This contribution is pretty simple, the type definition was using a syntax that isn't compatible with es6 (which is typescript's default syntax).

I am not sure why I didn't have this issue before, but noticed it on one of my build servers the type definition was failing as `export = SomeClass;` isn't really supported, instead it needs to use `export default SomeClass;`

Hope to see this change in trunk soon :)

### DIFF
--- a/src/js/index.d.ts
+++ b/src/js/index.d.ts
@@ -106,5 +106,5 @@ declare module "lottie-react-native" {
     reset(): void;
   }
 
-  export = AnimatedLottieView;
+  export default AnimatedLottieView;
 }


rodrigoelp:
No worries mate. I find lottie a fantastic tool :) wish I could collaborate a lot more. I haven't pushed some of the changes I've done in my private repo because I am still dealing with xcode 10 issues. But I will be doing a new PR soon removing the DevelopmentTeam configuration as those are reported as issues on the new build system.

emilioicai:
Sounds great! looking forward to that PR :)

src/js/index.d.ts
```diff
@@ -106,5 +106,5 @@ declare module "lottie-react-native" {
     reset(): void;
   }
 
-  export = AnimatedLottieView;
```
ljharb: This was valid TypeScript syntax which handles describing `module.exports =`, ftr.

rodrigoelp: Hi @ljharb, you are correct that `export =` is supported in typescript making it compatible with commonJS. Now, I provided this change as `export default` seems to match a lot better the intended use here and does not generate issues with the default tslint configuration.

rodrigoelp: @ljharb, my understanding might be correct and I will happy to correct this but, a default export is meant to address this, so the consumer of the code can use the following syntax:

```ts
import x from "module";
```

An `export =` forces the consumer to always consume the module/object as:

```ts
import x = require("module");
```

which is the commonjs way, not typescript's. In which case, is not the intended way.

ljharb: Typescript only works properly if you have esModuleInterop and synthetic imports enabled - with that, you can import it directly with the export = type syntax.

rodrigoelp: @ljharb could you elaborate a bit more? I am getting a bit confused.

What works properly in typescript when `"esModuleInterop": true` is set in the tsconfig.json file?

The usage of `export =` with which import declaration?

Or does the `import x from "module";` works properly only when `"esModuleinterop": true` is set?

I've just re-read  the typescript documentation and it specifies what I mentioned above. `export =` forces you to declare your import as `import x = require("module")` as opposed to `import x from "module"` stated in the documentation [here](https://github.com/react-community/lottie-react-native/blob/master/README.md#usage)

ljharb: If you’re using both of those settings, then you can import default *anything* - whether written in ESM or CJS - with a normal JS import.

rodrigoelp: Ah, I learnt something new, thanks for explaining!

In most of my projects I might be using the `esModuleInterop` off and I was trying to follow the documentation showing how to import the lottie view.

Either way, if the way the lottie view has been exported (as default) is causing issues, I am happy to create another pull request to rollback the change but I will suggest to include an addendum to the readme file about typescript and this caveat.

PR #393, opened 2018-10-11
Updated build.gradle file

SkariDamm:
Before change, this script caused a "Could not find method compileOnly()" and "Could not find method implementation()" when performing react-native run-android. (RN version 0.55.4)

### DIFF
--- a/src/android/build.gradle
+++ b/src/android/build.gradle
@@ -26,6 +26,6 @@ android {
 }
 
 dependencies {
-  compileOnly "com.facebook.react:react-native:+"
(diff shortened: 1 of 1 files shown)


emilioicai:
`provided` and `compile` will be deprecated later this year and therefore must be replaced by `compileOnly` and `implementation`. If you want your build script to work you might need to update gradle on your machine: https://stackoverflow.com/questions/51479691/how-can-i-upgrade-react-native-gradle-version

I would recommend gradle 3.1.2 or above

PR #395, opened 2018-10-15
Removed warning when compiling iOS project

rodrigoelp:
This change addresses a warning generated by xcodebuild as the deployment target of the base project was set up to 9.2 while the output library was set up to 7.0. I am just defaulting to the version of the OS configured in the parent project.

The definition of the parent property is [here](https://github.com/rodrigoelp/lottie-react-native/blob/b635f9c14f6fbb6c1bc0453e569fcc7b851b2c78/src/ios/LottieReactNative.xcodeproj/project.pbxproj#L181) and [here](https://github.com/rodrigoelp/lottie-react-native/blob/b635f9c14f6fbb6c1bc0453e569fcc7b851b2c78/src/ios/LottieReactNative.xcodeproj/project.pbxproj#L219)


rodrigoelp:
Hi @emilioicai, it is me again (I must be very annoying)... This commit fixes some of the warnings I've mentioned before whilst compiling iOS. Is really easy to forget some of these properties where set as part of the build phases and xcode does not really highlight these.

In any case, let me know if there are problems with the commit and I will fix it as soon as possible.

emilioicai:
Thanks @rodrigoelp for all your support :)

PR #402, opened 2018-10-30
Use commonjs object export compatible syntax

leighman:
The js file does not have a default export.
The typescript mechanism for a function or class commonjs `module.exports` is `export =`.

### DIFF
--- a/src/js/index.d.ts
+++ b/src/js/index.d.ts
@@ -124,5 +124,5 @@ declare module "lottie-react-native" {
     reset(): void;
   }
 
-  export default AnimatedLottieView;
+  export = AnimatedLottieView;
 }


leighman:
default exports are allowed but https://github.com/react-community/lottie-react-native/blob/master/src/js/index.js does not have one so it causes a runtime crash if you import it (as far as typescript is concerned) correctly.
One workaround is to use `esModuleInterop` but the correct seems to be to use `export =` as seen in the `export = and import = require()` section of the page you link to and on `DefinitelyTyped` for modules that have a single `module.exports` of a class or function.

emilioicai:
Yeah, that makes sense. Ok, I'm approving this

rodrigoelp:
Hi @emilioicai, I want to create an addendum to the documentation (or maybe, just dedicate a file to typescript and lottie's usage)

I've discussed with @leighman and others about this issue I introduced on a previous pull request, but I do want to make it easier for people using typescript to understand how to consume lottie.

PR #415, opened 2018-12-05
Add onAnimationFinish callback function

li-yuan2:
Add callback function `onAnimationFinish` which will be called when animation is finished. Note that this callback will be called only when `loop` is set to false.


Trancever:
@emilioicai Any ETA when this will land on npm?

PR #422, opened 2018-12-13
Removed "onAnimationFinish" from api.md

nicomontanari98:
Removed prop "onAnimationFinish" because not existing


emilioicai:
Hi @nicomontanari98 this is something coming in the next version of lottie-react-native

PR #427, opened 2019-01-04
feat: add dynamic theming

hamdiwanis:
This adds the ability to change animation colors in runtime using lottie dynamic properties.


kanobali:
This is great, now i can stop editing json files in runtime to do it . @emilioicai  @gpeal  what is blocking this from being merged ?

kanobali:
@emilioicai yes it worked for ios , for android i had to update android-lottie to 2.6.0 first

Codex-:
Lottie has no notion of a 'theme' and as such extensions to use these things should adhere more closely to the lottie style of consuming these properties. If a consumer wants to wrap constructs such as these then the consumer should handle these in their respective use cases and not have these forced upon them in the parent library. 

See the below for more details:
https://github.com/bradmartin/nativescript-lottie/pull/40

hamdiwanis:
hey @kanobali thanks for the gifs 👍

hamdiwanis:
hey @Codex- it's funny i also added this to nativescript and i responded to you there and as i said i value having a declarative API instead of having to reference all the elements to then theme them. but you are welcome to add an imperative one too. thanks 👍

emilioicai:
Closing this as theming is supposed to be handled natively in lottie-ios and lottie-android

robertherber:
This looks great (haven't tried it though). I don't understand why this isn't merged if it's working? If it's working it's better to have this possibility than to not have it (even though I guess @emilioicai means there's an even better solution).

Codex-:
@robertherber read the discussion linked above:
https://github.com/bradmartin/nativescript-lottie/pull/40

The notion of a theme does not exist in the parent libraries and should not have public api changes imposed upon them as the behaviour may change. Typically if you feel a need for the notion of a theme you would implement this in your own internal provider.

PR #455, opened 2019-03-14
basic compatibility for react-native .59

bondparkerbond:
This is a quick fix for basic react-native .59 support while a more robust solution with AndroidX and other fixes is being worked on.


bondparkerbond:
I tested this on both an iPhone SE on iOS 11 and Pixel XL on 8.0.0 with a scoped package in my work app to verify that this fix does what we want on RN 0.59 and the Lottie Animations still work. I did not notice any regressions. I also reverted changes made to publish a scoped version as they were not meant to go into this PR and were only for above mentioned validation purposes. This PR still needs a version bump before it is merged in and published if approved.

bondparkerbond:
@emilioicai Is there anything I can do to help get this approved and merged? It's my final blocker to finish upgrading to RN 0.59.1.

emilioicai:
Hey @bondparkerbond , my main concern is compatibility with older versions of RN. I guess we would need to modify the README to let devs know which version they should use depending on which version of RN they have. Could you add this to your PR?

Thanks!

bondparkerbond:
@emilioicai Sure! Are we adding this as a minor version bump or what version number should I reference this as? We could also reference this as an alpha/beta/RC version of a major version bump that I could pin to use and then you could release a new version once the full AndroidX support gets added so people could have access to the fix if they want it but don't have to worry about accidentally upgrading to it...

bondparkerbond:
@emilioicai I added RN version info to the README, bumped the lottie-react-native version number to 2.6.0 (which seems to be the most consistent option with how this package handles versioning), and updated the CHANGELOG. If anything needs fixing feel free to do so directly or let me know what you want changed. Alternatively, I think it should be ready to merge in if you think it's ok.

bondparkerbond:
@emilioicai Thanks for your help with this!

emilioicai:
@bondparkerbond thank you!

lorenc-tomasz:
Thanks for this, but we still need to find better solution that my dirty hack from https://github.com/react-native-community/lottie-react-native/issues/453 However one more time - big thanks as we can update our projects to RN 0.59.1

PR #459, opened 2019-03-26
Lock lottie-ios on version 2.5.0

lorenc-tomasz:
There are compilation errors with 2.5.3 from npmjs repository so we should lock lottie-ios on 2.5.0 until we will migrate to 3.x.x


bondparkerbond:
Will this be added to a new bug fix version once it gets merged?

emilioicai:
@bondparkerbond Yes, I will prepare a new version hopefully within this week

PR #471, opened 2019-04-15
Update lottie-android to 2.6.1 to work with cacheStrategy

stokesbga:
Android 28 and React native 0.59 works great with lottie 2.6.1, but 2.6 seems to break with cache strategy, even though its been deprecated.. This bumps the version to work with cache strategy


emilioicai:
@stokesbga please, check the PR we accepted a couple of weeks ago to have some context: https://github.com/react-native-community/lottie-react-native/pull/458

PR #485, opened 2019-05-17
upgrade to lottie-android 3.0.0

emilioicai:


example/android/app/build.gradle
```diff
@@ -137,14 +136,17 @@ android {
...
 }
 
 dependencies {
-    compile project(':lottie-react-native')
-    compile 'com.android.support:appcompat-v7:26.1.0'
-    compile 'com.android.support:support-annotations:26.1.0'
-    compile "com.facebook.react:react-native:+"  // From node_modules
-
+    implementation project(':lottie-react-native')
+    implementation 'androidx.legacy:legacy-support-v4:1.0.0'
```
rozPierog: Great idea with the update! But I have some reservations about updating RN libraries to `AndroidX`. 
I personally would love if everyone would switch to androidx all at once so we could avoid fragmentation as much as possible., but it's just not feasible. Right now 99% of RN libraries do not support android jetpack so adding dependency with androidx import is just gonna break about every app.
Can't help but notice that `lottie-react-native` uses `Support-v4` only for `ViewCompat.isAttachedToWindow()` in line https://github.com/react-native-community/lottie-react-native/blob/master/src/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManager.java#L118 and https://github.com/react-native-community/lottie-react-native/blob/master/src/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManager.java#L144
I propose to replace this call with https://developer.android.com/reference/android/view/View.html#isAttachedToWindow(), so it's gonna be `view.isAttachedToWindow`, and removing `support-v4` dependency altogether. Only downside to this is loosing compatibility with android apis 16-19 (so about 3.2% of Android devices https://developer.android.com/about/dashboards) but you gain compatibility with AndroidX _and_ Android Support Library projects, and thats huge!

influx6: Anychance of this, getting merged without the change in SDK? It seriously is causing issues.

emilioicai: @rozPierog I'm not sure if I get your suggestion. Could you send a PR with it so I can give it a try? Thanks!

changLiuUNSW:
Any chances to get this merged? I need this PR to fix my issue:
https://github.com/airbnb/lottie-android/issues/1252

emilioicai:
as @rozPierog mentioned, migrating to androidX could cause issues with a lot of other libraries. I'm reluctant to go ahead with this PR at this moment

example/android/app/build.gradle
emilioicai: @influx6 what kind of issues current `lottie-react-native` is causing at the moment? As I mentioned, I'm a bit hesitant to go ahead with this PR as per @rozpierog comment

rozPierog:
@emilioicai take a look at this: https://github.com/mikehardy/jetifier#module-maintainers

example/android/app/build.gradle
influx6: I also won't advise it due to androidx support libraries. We resolved our issues, as we realized we were exported in bodymovin with lottie3 format

dhee9000: @influx6 Your comment was incredibly helpful. Almost all the animations on the LottieFiles Community site were working but even the simplest scaling animations of mine weren't working. Turns out it's the setting you mentioned under Advanced Settings. Same solution as to @airbnb/lottie-web#1561 incase anyone else stumbles upon this. It looks like the documentation has been updated to say the current version requires AndroidX? This was confusing.

changLiuUNSW:
@emilioicai 
https://twitter.com/gpeal8/status/1132824596749537280
It looks like they have an one-off 3.0.3-support artifact that supports non-androidx apps

plouh:
Now that RN 0.60 is out, it would make sense to merge and release 3.0.0 -branch of lottie-react-native that would bring androidx support to the module and have 2.6.x branch exist while waiting for the community to adopt androidx more.

Javi:
As of RN 0.60, according to [the release blog post ](https://facebook.github.io/react-native/blog/2019/07/03/version-60)
> With this change, React Native apps will need to begin using AndroidX themselves.

zwenza:
Any news on this?

emilioicai:
AndroidX support is already present in release 3.0.4. I'm working on this but it has to be done in a separate PR as build.gradle has changed too much. I will keep this PR open until the other is on its way as a reference.

emilioicai:
closing in favor of https://github.com/react-native-community/lottie-react-native/pull/527

zwenza:
@emilioicai okay thanks!
My problem with 3.0.4 is https://github.com/airbnb/lottie-android/issues/1252 like @changLiuUNSW 
mentioned. So I have no issues with AndroidX but with this particular error that seems to be fixed in lottie-android 3

PR #487, opened 2019-05-24
Remove sub dependency

namvoeh:
# Summary
Currently, React Pod has been deprecated, and it also leads to conflict between `package.json` of Pod's React project and React project. 

* What issues does the pull request solve? Please tag them so that they will get automatically closed once the PR is merged

* What is the feature? (if applicable)
Remove sub dependency
* How did you implement the solution?
Just remove sub dependency from podspec.

* What areas of the library does it impact?
It affects iOS.

## Test Plan
There is not React dependency
image

NO

1. Install lottie-react-native lib
2. `cd ios/ && pod install`
3. Run react native after all
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/lottie-react-native.podspec
+++ b/lottie-react-native.podspec
@@ -15,6 +15,5 @@ Pod::Spec.new do |s|
   s.source       = { :git => "https://github.com/react-community/lottie-react-native.git", :tag => "v#{s.version}" }
   s.source_files  = "src/ios/**/*.{h,m}"
 
-  s.dependency 'React'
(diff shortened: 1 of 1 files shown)


namvoeh:
I found that we can add React dependency from podfile as https://facebook.github.io/react-native/docs/integration-with-existing-apps.html#configuring-cocoapods-dependencies

PR #507, opened 2019-06-28
Fix xcodeproj

radex:
# Summary

This fixes the Xcode project. https://github.com/react-native-community/lottie-react-native/pull/486 broke it — it made this library work only via CocoaPods. This PR fixes it. Unfortunately the upgrade to Lottie iOS v3 also seems to mean iOS 12 is required if you're linking via Xcode project

## Test Plan

Create app, link "traditionally" lottie-react-native and lottie-ios, see if compiles

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |

## Checklist

- [ ] I have tested this on a device and a simulator


emilioicai:
thanks @radex

enagorny:
@radex Could you please explain what is required to make it work on iOS 11 and lower?
Do we need to update readme to reflect it?

radex:
@enagorny 

> Apps that use static Xcode project linking need to set iOS deployment version to iOS 12 *or* switch to CocoaPods-based linking (using frameworks) *or* downgrade `lottie-react-native` to version **_2.6.1_**.

iOS 12 requirement is from `lottie-ios v3` when using static libraries -- you can take up the issue there

Issue #521, opened 2019-07-22
Version 3.0.3 on Android, RN 0.60.0 crashes at start.


ricbermo:
### Description

The Android application is crashing just after it starts, meaning it builds and launches successfully but nothing else.

### Steps to Reproduce

1. create a new repo
2. Install the package
3. run `react-native run-androd`

**Expected behavior:**
The app should render the animtions

**Actual behavior:**
The app crashes

**Versions:**

RN: 0.60.0
Lottie: 3.0.3

** Repo **
https://github.com/ricbermo/open-source-contrib

sathindukavneth:
Had the same issue after updating into 3.0.3

roninprogrammer:
Hi guys , 

I believe this is because the library didnt auto-linked properly to native code ,

In order to solve this , please add this manually: 

**MainApplication**

```
- add import com.airbnb.android.react.lottie.LottiePackage;
- add packages.add(new LottiePackage())in List<ReactPackage> getPackages();
```

**Build.gradle**
`add implementation project(':lottie-react-native')`

**Settings.gradle**
```
include ':lottie-react-native'
project(':lottie-react-native').projectDir = new File(rootProject.projectDir, '../node_modules/lottie-react-native/src/android')

```

emilioicai:
thanks a lot @roninprogrammer. That was actually the issue. I will add this to the README until autolinking is fixed.

roninprogrammer:
> Had the same issue after updating into 3.0.3

@ricbermo @sathindukavneth you may try this manual implementation

ricbermo:
@emilioicai @roninprogrammer thanks for your help. I can confirm that workarounds the issue.

ifsnow:
It works well without crash in the recently updated 3.0.4

roninprogrammer:
> It works well without crash in the recently updated 3.0.4

Awesome

roninprogrammer:
> @emilioicai @roninprogrammer thanks for your help. I can confirm that workarounds the issue.

you may close the issue =)

PR #524, opened 2019-07-23
Fix crash on Android and RN 0.60

ifsnow:
Related to : #521 

This config makes `autolink` work properly.

I hope this helps.


emilioicai:
Thanks a lot @ifsnow !! this was much needed :)

PR #548, opened 2019-09-04
Fix loading assets from app bundle on iOS

janicduplessis:
# Summary

The method for loading named assets on iOS wasn't actually doing anything. I added a simple implementation using AnimatedView(name:) which does exactly what we want and also takes care of caching the loaded animation.

None

- Add an animation json to your xcode project
- Add a lottie view like: `<LottieView source="nameOfAnimationWithNoExtension" />

## Compatibility

N/A

## Checklist


emilioicai:
@janicduplessis thanks! looking good to me

PR #585, opened 2019-11-01
implement reanimated support

JoaoLSS:
# Summary

Add support for using Reanimated Nodes in progress prop

## Test Plan

Should run exactly the same since the only change is it renders a Reanimated Component if the progress prop is a Reanimated Node

Use Reanimated.Node in progress prop

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] I updated the typed files (TS and Flow)


src/js/index.d.ts
```diff
@@ -34,7 +35,7 @@ declare module "lottie-react-native" {
      * animation will correspondingly update to the frame at that progress value. This
      * prop is not required if you are using the imperative API.
      */
-    progress?: number | Animated.Value;
+    progress?: number | Animated.Value | Reanimated.Node;
```
mbret: This type does not compile with typesscript since Node require an argument

`Generic type 'AnimatedNode<T>' requires 1 type argument(s).`.

emilioicai: @mbret good catch! could you submit a PR to fix it?

dozoisch: @emilioicai this pr also forces you to have react-native-reanimated install, as otherwise errors because the module is not resolved. I'm not sure if that was an intended side-effect

dozoisch: I think it's possible to change the import to make it conditional, but it requires a little bit more handling. Or maybe we export 2 different components. One that imports reanimated and one that uses the other one.

emilioicai: I think you can make it using `require` directly but I'm afraid reanimated would still need to be added as a dependency which may confuse users who doesn't use reanimated on their projects. I'm happy to think this over though as I consider Reanimated is a great tool.

alexichristakis:
why was this reverted?

dozoisch:
@alexichristakis see https://github.com/react-native-community/lottie-react-native/pull/585#discussion_r346600256 ! It was causing issues when people didn't have react-native-reanimated installed

NTag:
Any plan to add it but without causing issues when people don't have react-native-reanimated? It would be awesome 🙏

JoaoLSS:
sorry for the bad MR, I forget to delete Reanimated from the dependencies of the main project as I was testing, so I didn't get the complain. I don't think reanimated has a too small use, because of its use in React Navigation, so synchronized animations as I was trying to accomplish need this implementation, @NTag MR seems promising though

davx1992:
Hi is there any update on this?

NTag:
I have some comments to address on my PR, I'll do it as soon as possible.

PR #607, opened 2020-01-24
Use react-native-reanimated if present

NTag:
# Summary

Add support for react-native-reanimated. Contrary to https://github.com/react-native-community/lottie-react-native/pull/585 it's not necessary to have react-native-reanimated installed (it's not added as a peer dependency). We try to import it, and if it succeed, then we test if `progress` is a reanimated node, and if that's the case we create a reanimated component (most of the code comes from @JoaoLSS in the related MR).

## Test Plan

It should work as previously in applications without react-native-reanimated.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

- [x] I updated the typed files (TS and Flow)

### DIFF
--- a/src/js/index.d.ts
+++ b/src/js/index.d.ts
@@ -39,7 +39,7 @@ declare module "lottie-react-native" {
      * animation will correspondingly update to the frame at that progress value. This
      * prop is not required if you are using the imperative API.
      */
-    progress?: number | Animated.Value;
+    progress?: number | Animated.Value | any;
 
     /**
      * The speed the animation will progress. This only affects the imperative API. The
--- a/src/js/LottieView.js
+++ b/src/js/LottieView.js
@@ -8,6 +8,12 @@ import {
   StyleSheet,
   ViewPropTypes,
 } from 'react-native';
+let Reanimated;
+try {
+  Reanimated = require('react-native-reanimated').default;
+} catch (e) {
+  // react-native-reanimated isn't installed, we'll fallback on React.Animated
+}
 import SafeModule from 'react-native-safe-modules';
 import PropTypes from 'prop-types';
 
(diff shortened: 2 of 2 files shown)


src/js/index.d.ts
```diff
@@ -39,7 +39,7 @@ declare module "lottie-react-native" {
      * animation will correspondingly update to the frame at that progress value. This
      * prop is not required if you are using the imperative API.
      */
-    progress?: number | Animated.Value;
+    progress?: number | Animated.Value | any;
```
JoaoLSS: perhaps do the same as the try import on the js file,
type AnimatedSource = Animated.Value
try {
 Reanimated = require(...).default
AnimatedSource = Animated.Value | Reanimated.Node
}

dont know if is gonna work though, I'm seen all of this today after my MR haha

JoaoLSS: turns out you can't reassign a type, but since the class implementation is so basic (https://github.com/software-mansion/react-native-reanimated/blob/master/react-native-reanimated.d.ts) it could be defined within the d.ts file to be equal to the original one, since typescript only cares about the same implementation, and not which class is which.
something like:

interface ReanimatedNode {
 new (nodeConfig: object, inputNodes?: ReadonlyArray<AnimatedNode>): ReanimatedNode
isNativelyInitialized(): boolean
' __value': any
}

should suffice

JoaoLSS:
I have tried your approach, but seems RN doesnt deal with requires exactly as node does, https://github.com/facebook/react-native/issues/5079, so RN still throws :/

emilioicai:
closing as now we have a better typing for this

PR #684, opened 2020-09-18
fix: Xcode 12 compatibility

Fetten:
# Summary

The latest Xcode 12 fails to build when a module does not depend on `React-Core` directly. This change is necessary for all native modules on iOS. For details please see: https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

## Test Plan
1. Build an App with Xcode 11.x. Result: The app builds.
2. Build an App with Xcode 12. Result: The app will not build when depending on `React`.
3. Build an App with Xcode 12 with the PR changes applied. Result: The app builds.

macOS with Xcode 12

see Test Plan

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/lottie-react-native.podspec
+++ b/lottie-react-native.podspec
@@ -15,6 +15,6 @@ Pod::Spec.new do |s|
   s.source       = { :git => "https://github.com/react-community/lottie-react-native.git", :tag => "v#{s.version}" }
   s.source_files  = "src/ios/**/*.{h,m,swift}"
   s.swift_version = "5.0"
-  s.dependency 'React'
+  s.dependency 'React-Core'
   s.dependency 'lottie-ios', '~> 3.1.8'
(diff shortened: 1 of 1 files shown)


radko93:
Lottie is a Swift library, so you should check if this problem does not occur https://github.com/facebook/react-native/issues/29633#issuecomment-694700346

Fetten:
> Lottie is a Swift library, so you should check if this problem does not occur [facebook/react-native#29633 (comment)](https://github.com/facebook/react-native/issues/29633#issuecomment-694700346)

You are right, @radko93, I re-checked my PR. Thought we updated and tested the change as well in our latest project, but it looks like I confused some package here. Closing this.

PR #685, opened 2020-09-22
Fix React dependency so XCode 12 can compile

harunsmrkovic:
# Summary

Fixes issue described in https://github.com/facebook/react-native/issues/29633

No tests needed, it was a wrong dependency.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/lottie-react-native.podspec
+++ b/lottie-react-native.podspec
@@ -15,6 +15,6 @@ Pod::Spec.new do |s|
   s.source       = { :git => "https://github.com/react-community/lottie-react-native.git", :tag => "v#{s.version}" }
   s.source_files  = "src/ios/**/*.{h,m,swift}"
   s.swift_version = "5.0"
-  s.dependency 'React'
+  s.dependency 'React-Core'
   s.dependency 'lottie-ios', '~> 3.1.8'
 end


mburnet6-ibm:
@harunsmrkovic I tried applying this locally and got an error during `pod install`.

```
[!] The following Swift pods cannot yet be integrated as static libraries:

The Swift pod `lottie-react-native-iOS` depends upon `React-Core-iOS`, which does not define modules. To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries), you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies.
```

can you share how you have your Podfile setup to make this work?

harunsmrkovic:
Hm, it could be that `use_frameworks!` gets rids of this error for me... Not sure, haven't seen that error in awhile...

emilioicai:
`lottie-react-native` is a Swift module, if we merge this PR we will start having this issue: https://github.com/facebook/react-native/issues/29633#issuecomment-694700346

We are waiting for a change in React Core's podspec to fix the issue: https://github.com/facebook/react-native/pull/29995

emilioicai:
Looks like the change in React Core is already merged. Can anybody help me here testing this before merging?

emilioicai (approved):

emilioicai:
This has been resolved on https://github.com/lottie-react-native/lottie-react-native/pull/738

PR #701, opened 2020-11-11
[Infrastructure] Reduce npm package size

retyui:
**To test:**
```sh
npm pack --dry-run
```

**Diff:**
|   | package size | unpacked size |
|---|---|---|
|Before| 79.6 kB | 156.1 kB |
|After| 78.6 kB | 154.0 kB |

- [See package content on the web](https://unpkg.com/browse/lottie-react-native/)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -41,3 +41,7 @@ book.json
 .eslintrc
 .eslintignore
 .yarnrc
+babel.config.json
+.flowconfig
+Gemfile
+Gemfile.lock


.npmignore
```diff
@@ -41,3 +41,7 @@ book.json
 .eslintrc
 .eslintignore
 .yarnrc
+babel.config.json
+.flowconfig
+Gemfile
+Gemfile.lock
```
ljharb, line 47: By removing these, you're further breaking the ability to `npm explore foo && npm install-test`, which doesn't seem worth the mere 1k/4k size decrease (the eslint and yarn config files should be included too, ofc)

retyui: Never used `npm explore` and never heard that someone use it)

ljharb: It's why every npm package I solely maintain, or create, also publishes all its tests to npm.

emilioicai: I'm with @ljharb here, not sure if the size improvement is worth this PR

PR #710, opened 2020-12-04
Updated resizeMode to use FIT_CENTER instead of CENTER_INSIDE.

collinprice:
When using `resizeMode="contain"` on Android the lottie image does not grow to fill the container. If the actually dimensions of the lottie are 100x100 and the container is 50x50 then `resizeMode="contain"`will make the lottie shrink to 50x50 but if the container is 200x200 the lottie does not grow past 100x100.

Changing `ImageView.ScaleType.CENTER_INSIDE` to `ImageView.ScaleType.FIT_CENTER` resolves this issue.


matinzd:
Can you rebase your PR with the latest changes on the master branch?
Thanks for your contribution!

PR #714, opened 2020-12-31
C++/WinRT implementation

aschultz:
Wraps the WinUI 2.x AnimatedVisualPlayer to add support for lottie-react-native in React Native Windows apps. 

This implementation uses dependency injection through `ILottieSourceProvider` to allow the consuming app to choose how sources are resolved. The example app includes animation .idl/.h/.cpp/.cs files generated by LottieGen 7.x, which are then mapped to names that can be used in JS.

- Supports both C++/WinRT and C# UWP apps. Example projects included for each.
- Supports compile-time generation of animations for both languages (reference your animations in JS by name)
- Supports dynamic loading of JSON at runtime for C# apps
- Codegen is streamlined through MSBuild. Once you have your project set up, you can just add new JSON files as `LottieAnimation` items in your project. This system is opt-out, so you can roll your own loader if you want.

TODO (future PR):
- Extract framerate from dynamic sources to support play(from,to) with correct time mapping. Out of scope for this PR.


aschultz:
@jonthysell Curious about your take on this since I see you have a C# implementation in the works.

src/windows/cppwinrt/LottieAnimationViewManager.cpp
```diff
@@ -0,0 +1,174 @@
...
+                    }
+                    control.ResizeMode(stretch);
+                }
+                else if (propertyName == "sourceName") {
+                    auto sourceName = ReadValue<std::optional<std::wstring>>(propertyValue);
+                    auto source = m_lottieSourceProvider.GetSourceFromName(sourceName.value_or(L""));
+                    control.Source(source);
+                }
+                else if (propertyName == "sourceJson") {
+                    auto sourceName = ReadValue<std::optional<std::wstring>>(propertyValue);
```
aschultz, line 105: Rename var to `sourceJson`

aschultz:
I figured out how to convert from frames to normalized time using some of the new capabilities in LottieGen. This avoids the need for the hard-coded framerate value when using codegen'd sources.

Given that its a bit convoluted to ensure everything is set up properly, I also set up MSBuild targets to automate everything. Now all you need to do is add `<LottieAnimation>` items to your project pointing to your JSON files and import `LottieGen.Cpp.targets`. It will take care of running LottieGen and adding the generated files to the build. It also generates an `ILottieSourceProvider` automatically to make it trivial to set up the React component.

TODO:
- Integrate LottieGen NuGet package. We need version 7, which isn't released yet, so I'm using a locally built copy of the exe right now.
- Change the generated source provider to be a full WinRT type so it doesn't require a weird *.g.cpp inclusion.
- Improve the Example to cover more functionality (like Stop and Play(start, end))

src/windows/cppwinrt/ReactPackageProvider.idl
```diff
@@ -0,0 +1,21 @@
...
+        void SetScalarProperty(String propertyName, Double value);
+    };
+
+    interface ILottieSourceProvider
+    {
+        Microsoft.UI.Xaml.Controls.IAnimatedVisualSource GetSourceFromName(String name);
+        Microsoft.UI.Xaml.Controls.IAnimatedVisualSource GetSourceFromJson(String json);
+    }
+
+    [default_interface]
```
aschultz, line 16: Remove the default_interface and mark IReactPackageProvider as default.

src/windows/cppwinrt/LottieAnimationViewManager.cpp
```diff
@@ -0,0 +1,178 @@
+#include "pch.h"
+#include "LottieAnimationViewManager.h"
+
+#include <winrt/Windows.UI.Xaml.h>
+#include <winrt/Windows.UI.Xaml.Controls.h>
+#include <winrt/Windows.UI.Xaml.Media.h>
+#include <winrt/Microsoft.UI.Xaml.Controls.h>
```
asklar, line 7: i like the direction this is going in! One thing to consider is that we are working on moving react native for windows on top of WinUI 3. For the time being we still have system xaml customer, obviously, so ideally we should get both scenarios working. The way we have done this in the core framework is by using some #defines and namespace aliases, e.g.
```
#ifdef USE_WINUI3
namespace xaml = winrt::Windows::UI::Xaml;
// ...
#else
namespace xaml = winrt::Microsoft::UI::Xaml;
// ...
#endif
```
(see other examples in the RNW code).
Would it be possible for you to do something similar in this project so that RNW apps that want to adopt WinUI3 are able to do so?

src/windows/cppwinrt/LottieReactNative.idl
```diff
@@ -0,0 +1,43 @@
...
+        LottieView();
+        LottieView(Microsoft.ReactNative.IReactContext context, ILottieSourceProvider sourceProvider);
+
+        Windows.UI.Xaml.Media.Stretch ResizeMode { get; set; };
+
+        void SetSourceName(String name);
+        void SetSourceJson(String json);
+        void SetLoop(Boolean loop);
+        void SetProgress(Double progress);
+        void SetSpeed(Double speed);
```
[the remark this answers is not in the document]
aschultz: Most aren't applied immediately, but instead queued up to take effect at varying times based on load state. As such, its not clear what you'd want as the read value -- simply what was last passed in, or the actual state at that moment? Since there's nothing that needs to read them right now, I figured any decision there could be made when there's a scenario that needs the info.

src/windows/cppwinrt/LottieReactNative.vcxproj
```diff
@@ -0,0 +1,173 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="LottieReactNative.def" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
+  <PropertyGroup Label="ReactNativeWindowsProps">
+    <ReactNativeWindowsDir Condition="'$(ReactNativeWindowsDir)' == ''">$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), 'node_modules\react-native-windows\package.json'))\node_modules\react-native-windows\</ReactNativeWindowsDir>
```
jonthysell, line 146: You should replace `$(MSBuildThisFileDirectory)` with `$(SolutionDir)` to ensure this module plays well with other modules nad/or non-standard app setups.

src/windows/cppwinrt/packages.config
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.UI.Xaml" version="2.3.191129002" targetFramework="native" />
```
jonthysell, line 3: So the module itself doesn't depend on LottieWindows?

aschultz: Correct. It's up to the app to implement an ILottieSourceProvider and inject it into the module during setup. That allows the app to choose how to convert names or JSON into IAnimatedVisualSource, which could be through codegen, through LottieVisualSource, or some custom implementation.

The targets files greatly simplify this setup by auto-generating a provider and including it in the app's build (though you still need to new it and pass it into the module). For C++ apps, this provider only supports codegen and returns null for JSON requests. For C# apps, it supports both (using LottieVisualSource for JSON).

For codegen, LottieGen either needs to be installed globally on the system, or an MSBuild variable set to give its location. For LottieVisualSource, the app will need to add Lottie-Windows as a dependency.

The targets files have some initial documentation in them, but I can move it out into the README and add a section for Windows setup.

src/windows/cppwinrt/LottieAnimationViewManager.cpp
aschultz: Yeah, I can do that. What's the mechanism for surfacing the USE_WINUI3 flag to the consuming application? Should I be importing a specific property sheet into this library or is it taken care of by Microsoft.ReactNative.Uwp.CppLib.props / Microsoft.ReactNative.Uwp.CppLib.targets?

asklar: thanks! the way we did it for RNW core was:
1. A build-time property called `UseWinUI3`. This is passed in by the build command line (or alternatively environment variable) when invoking msbuild (`msbuild /p:UseWinUI3=true ...`)
2. The vcxproj includes `$(ReactNativeWindowsDir)\PropertySheets\WinUI.props` (via `Microsoft.ReactNative.Uwp.CppLib.props`). This defines some variables and pushes a compiler preprocessor definition for `USE_WINUI3` so that code can then switch on `#ifdef USE_WINUI3`

namrog84:
Hey any traction on this?  We'd love to see this.

src/windows/cppwinrt/PropertySheets/LottieGen.CSharp.targets
```diff
@@ -0,0 +1,216 @@
...
+        We're not using incremental build here because changes to item metadata (like the Name attribute) wouldn't trigger rebuild.
+        Instead, we can use WriteOnlyWhenDifferent to avoid changing the output file if our code is the same.
+    -->
+    <Target Name="LottieGen_CreateProvider" 
+        DependsOnTargets="LottieGen_CreateOutputPath;LottieGen_UpdateMetadata"
+        Outputs="$(_LottieProviderFilePath)">
+        <PropertyGroup>
+            <_UseLottieVisualSource Condition="'$(EnableLottieDynamicSource)' == 'true'">#define LOTTIEVISUALSOURCE</_UseLottieVisualSource>
+            <_LottieProviderContent>
+<![CDATA[
```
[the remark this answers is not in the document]
aschultz: I moved some of the static code out to a separate file, which also helps avoid red squiggles in VS. Still need the codegen'd file for listing out all the codegen'd files though.

aschultz:
@jonthysell @asklar I think this is in a good state for consumption now.

I worked with the LottieGen devs to get the tool built as a portable .NET app with MSBuild bindings, so it can now be easily installed into the consuming app as a NuGet package and made available to the codegen targets. No need for .NET core installations or other dependencies. It also generates more useable code now that actually exposes animation metadata.

I streamlined the RN setup process to minimize changes needed to project files and try to permit flexibility for WinUI versioning. The readme is updated with setup steps.

Both C# and C++ example apps are functional. By default, C++ gets access to codegen, C# to both codegen and dynamic loading via LottieVisualSource. The ReactPackageProvider accepts an ILottieSourceProvider object, so in theory if an app wanted to provide some other way of loading the JSON it could implement that interface and plug in.

The React side of the example doesn't do a great job showing off the capabilities of the player, but fixing it requires pulling in a community Slider control and I don't have the resources to verify those changes on iOS/Android.

asklar (approved):

aschultz:
@jonthysell @emilioicai Friendly ping. Sounds like everything looks good? I merged main and updated everything to work with RNW 0.66.

I can put out a followup PR to remove the other Windows implementation.

emilioicai:
Hey @aschultz thanks a lot for your work, I'm looking forward to merging this one in but I would need some assistance in testing it in a windows machine. @jonthysell could you give a hand with this?

jonthysell (approved):
Looks good to me, both examples (csharp and cppwinrt) build and run just fine x86/x64.

aschultz:
@emilioicai Just waiting on you. I merged main in again, added a few doc comments, and fixed a bug in the example JS code, but otherwise nothing's changed since jonthysell signed off.

emilioicai:
Thanks @aschultz , I'm doing some testing over the weekend and once it's done I will merge and prepare a new version with it. Again, thanks a lot for taking this one, I reckon there's a lot of effort in this PR.

emilioicai (approved):

emilioicai:
This is merged and will be released on v6.0.0

Thanks @aschultz  and @jonthysell

aschultz:
@emilioicai Thanks! Any idea when v6 will release?

PR #735, opened 2021-03-10
fix: move `lottie-ios` to `peerDependencies`

maxkomarychev:
`lottie-ios` is a project with native code and it has to be linked.

react-native autolinking can only work as long as project is listed in
the output of `react-native config` which considers `package.json` in
order to build list of dependencies which need to be linked.

currently having `lottie-ios` in `dependencies` does not make lots of
sense since in order to successfully install and use
`lottie-react-native` one has to install `lottie-ios` separately and
make sure it's listed in `package.json`.

in fact it may lead to situatoin when two copies of `lottie-ios` may be
installed in case if range of `lottie-ios` inside package.json of
`lottie-react-native` and inside actual package of the app are
conflicting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,11 +43,11 @@
   "homepage": "https://github.com/airbnb/lottie-react-native#readme",
   "peerDependencies": {
     "react": "*",
-    "react-native": ">=0.46"
+    "react-native": ">=0.46",
+    "lottie-ios": "^3.1.8"
   },
   "dependencies": {
     "invariant": "^2.2.2",
-    "lottie-ios": "^3.1.8",
     "prop-types": "^15.5.10",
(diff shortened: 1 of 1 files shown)


emilioicai:
This looks good to me but I guess the README should be updated accordingly. @maxkomarychev would you be able to update it?

maxkomarychev:
@emilioicai I can but from what I've seen it looks like it already suggests to install `lottie-ios` explicitly (screenshots below), can you suggest another place where it would make sense to tell this? 🤔 

Screenshot 2021-03-11 at 20 19 25
Screenshot 2021-03-11 at 20 19 21

emilioicai:
Actually you are right, no need to explain this in other places. Thanks!

emilioicai (approved):

PR #771, opened 2021-05-16
Swift Version on Pod spec

marcosrdz:
Fixes this error during RN macOS init:

[!] Unable to determine Swift version for the following pods:  - `lottie-react-native-macOS` does not specify a Swift version and none of the targets (`bluewallet-macOS`) integrating it have the `SWIFT_VERSION` attribute set. Please contact the author or set the `SWIFT_VERSION` attribute in at least one of the targets that integrate this pod.


emilioicai (approved):

LeonardoRick:
I'm using `"lottie-react-native": "4.0.2",` and ` s.swift_version = "5.0"` is not on my `podspec`. Was it really merged afterall?

PR #773, opened 2021-05-20
Added support for dynamic text

MrGVSV:
# Reasoning
To make localizations easier to create for Lottie animations featuring text, without having to parse through the source JSON manually.

# Platform Differences
The implementations for changing the text of an animation differ between iOS and Android (as far as I could tell). Therefore, two props were added, one for each platform. This is not ideal, obviously, but it allows for the basic functionality without having to edit the native repos.

There may be some other way to overcome these differences that I wasn't able to find. If there is, that would be a much cleaner solution.

## iOS
For iOS, the implementation is done using a [`DictionaryTextProvider`](https://github.com/airbnb/lottie-ios/blob/79a0b70547f7c40ea54c67487f935fa2f2eaaadc/lottie-swift/src/Public/TextProvider/AnimationTextProvider.swift). This takes in a KeyPath to a text layer (though in testing it seems to only respond to the actual layer name) and sets the text value of that layer.

```js
// Example prop usage (iOS)
textFiltersIOS={[
    {
        keypath: "My Lottie Layer",
        text: "New Text"
    }
]}
```

## Android
For Android, the implementation is different. The [`TextDelegate`](https://github.com/airbnb/lottie-android/blob/90b37098b6bccda371bf0a1e810ce1611959284d/lottie/src/main/java/com/airbnb/lottie/TextDelegate.java) class takes in a text _value_ and replaces its occurrences throughout the animation (in other words, it does not perform the same KeyPath search that iOS does).

```js
// Example prop usage (Android)
textFiltersAndroid={[
    {
        find: "Old Text",
        replace: "New Text"
    }
]}
```

# Changes
- [x] Implemented dynamic text support
- [x] Updated docs
- [ ] Added example to example project (see note below)

<hr />

> **Note**: Unfortunately, I couldn't get the example project up and running on my machine for some reason. Instead, I used one of my own projects to test the changes.


mashrukh-islam:
@emilioicai I am working on a project where I need support for dynamic text using lottie-react-native. Is this PR mergeable?

innoc:
I am working a project that also requires dynamic text change. Can we get this merged pls 🥺

Maxime-Robinet:
That would be amazing. Is this mergeable, or is there any way in which we could help?

MrGVSV:
> That would be amazing. Is this mergeable, or is there any way in which we could help? 

I’m not too sure why this hasn't been merged yet, but maybe it has to do with the missing example?

I had trouble getting the example project to work on my machine, so if anyone has the project working and would like to help add in an example showcasing this feature, that would be great! 😃

Hopefully then, this could be fully merged?

emilioicai (approved):

Rag0n:
Now it's merged, but new version is still not released.

emilioicai:
hopefully a new version will be released next week

PR #795, opened 2021-08-05
Fix `onAnimationFinish` callback behavior after resuming on iOS

CatalinVoss:
This PR fixes an issue that causes erratic `onAnimationFinish` callbacks on iOS when we call `resume()`. Since the callback was passed in with the call to `play(..., completion: handler)`, and since `resume()` calls `play()` without any arguments, the animation completion callback gets lost if we ever pause the animation. (Note: that's not what's done in the android version), 

Here, we remove callback the references from the iOS manager module. Since `ContainerView` holds the property, we can always pass the callback directly in there.

### DIFF
--- a/src/ios/LottieReactNative/ContainerView.swift
+++ b/src/ios/LottieReactNative/ContainerView.swift
@@ -84,7 +84,7 @@ class ContainerView: RCTView {
 
     func play(fromFrame: AnimationFrameTime? = nil, toFrame: AnimationFrameTime) {
         let callback: LottieCompletionBlock = { animationFinished in
-            if let onFinish = view.onAnimationFinish {
+            if let onFinish = self.onAnimationFinish {
                 onFinish(["isCancelled": !animationFinished])
             }
         }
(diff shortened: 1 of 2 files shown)


src/ios/LottieReactNative/ContainerView.swift
```diff
@@ -82,14 +82,26 @@ class ContainerView: RCTView {
         applyProperties()
     }
 
-    func play(fromFrame: AnimationFrameTime? = nil, toFrame: AnimationFrameTime, completion: LottieCompletionBlock? = nil) {
+    func play(fromFrame: AnimationFrameTime? = nil, toFrame: AnimationFrameTime) {
+        let callback: LottieCompletionBlock = { animationFinished in
+            if let onFinish = view.onAnimationFinish {
```
Akeyo21, line 87: `view.onAnimationFinish` should be `self.onAnimationFinish`. Same as the one in `play()`

Akeyo21 (commented):
Aside from the comment I added, everything is okay. Thank you adding this in @CatalinVoss

src/ios/LottieReactNative/ContainerView.swift
CatalinVoss: Yep, forgot to move that over. Great catch!

CatalinVoss:
@emilioicai let us know what you think!

emilioicai:
Looks good to me, thanks @CatalinVoss !

emilioicai (approved):

PR #799, opened 2021-08-12
migrated to gradle 7

andresesfm:


emilioicai:
Thanks a lot Andres, this is very handy

andresesfm:
> Thanks a lot Andres, this is very handy

Glad to be of help 👍

PR #824, opened 2021-12-07
Fix UITraitCollection.hasDifferentColorAppearance check for tvOS version less than 13.0

jhen0409:
I got the following build error on tvOS after upgrade lottie-react-native to v5:

```
'hasDifferentColorAppearance(comparedTo:)' is only available in tvOS 13.0 or newer
```

This simple fix just add '#available' version check for tvOS.


jhen0409:
@emilioicai can you review this? Thanks!

emilioicai (approved):

emilioicai:
Thanks!

PR #839, opened 2022-01-26
fix: tvOS <13.0 build issue

motorosgep:
This PR fixes build issues on tvOS <13.0.

The error was 
`‘hasDifferentColorAppearance(comparedTo:)’ is only available in tvOS 13.0 or newer`
caused by a missing `#available` guard condition for tvOS.


matinzd:
Please update your PR with the latest changes and let us know.

PR #844, opened 2022-02-15
Allow user press through lottie animation when pointerEvents='none'

sdg9:
React Native supports [pointerEvents](https://reactnative.dev/docs/view#pointerevents) but they don't work as expected with `LottieView`.

## Scenario
```jsx
<View>
  <LottieView 
    pointerEvents="none"
    style={{ position: 'absolute', zIndex: 1, width: 500 }}
    source={require('./someAnimation.json')}
    autoPlay={true}
  />
  <TextInput placeholder="Type here" />
</View>
```
- `LottieView` is absolutely positioned with zIndex of 1 so that animations play on top of existing content.  
- The `pointerEvents="none"` prop is added to `LottieView` so that lower zIndex components can receive touch events.
- It does not work as expected as the outermost `View` of `LottieView`'s implementation doesn't receive the prop.

## Proposed Solution
Let the outermost `View` receive the `pointerEvents` prop.

#### Before
NoTyping
zIndex 0 never receives touch event.

#### After
Typing
zIndex 0 receives touch event as expected.

### DIFF
--- a/src/js/LottieView.js
+++ b/src/js/LottieView.js
@@ -154,7 +154,7 @@ class LottieView extends React.PureComponent {
   }
 
   render() {
-    const { style, source, autoSize, ...rest } = this.props;
+    const { style, source, autoSize, pointerEvents, ...rest } = this.props;
 
     const sourceName = typeof source === 'string' ? source : undefined;
     const sourceJson = typeof source === 'string' ? undefined : JSON.stringify(source);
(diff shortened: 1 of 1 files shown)


goxr3plus:
@sdg9 @emilioicai   We really need this too 😃 , because we are facing production problems because of pointerEvents not working with LottieView

Miigaarino:
Need this ++

devpgcs:
Hey guys I hope this work for you as it work for me:

```
const SomeComponent = () => {
    return (
        <>
            ... Some other components
            <View pointerEvents="none" style={{ width: "100%", height: "100%", position: "absolute" }}>
                <LottieView autoPlay source={require('../the-animations-folder/your-animation.json')} />
            </View>
        </>
    )
}
```

So, you just need to wrap it into a `View` which should cover the whole screen and which must have the `pointerEvents="none"`

goxr3plus:
.

matinzd:
Would you please rebase your branch with the master?

PR #855, opened 2022-03-14
fix: allow one to change lottie props and keep the animation running

cabelitos:
If one tries to update colorFilters while the animation is running,
prior to this patch the animation would simple stop running.
This is caused because applyProperties() always sets
animationView.currentProgress, which according to the
docs stops the animation in case it's running.

In order to prevent such scenario the applyProperties()
will check if the animation is running before any updates.
After all props are updated, it will resume the animation in
case it was playing.

https://airbnb.io/lottie/#/ios?id=current-progress


emilioicai (approved):

cabelitos:
Thank you!

PR #863, opened 2022-04-24
Hotfix/view prop types

dalollio:
It's a simple hotfix changing the ViewProTypes imports, to suppress warning log.

### DIFF
--- a/src/js/LottieView.js
+++ b/src/js/LottieView.js
@@ -6,7 +6,6 @@ import {
   View,
   Platform,
   StyleSheet,
-  ViewPropTypes,
   requireNativeComponent,
   NativeModules,
   processColor,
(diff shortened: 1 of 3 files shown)


thespacemanatee:
@emilioicai could we please release a new minor version with this change? Patch-package is not working as the library is using the files in the /lib directory and this issue is causing the dev environment to be bogged down by logs. Thank you!

src/js/LottieView.js
```diff
@@ -6,11 +6,11 @@ import {
   View,
   Platform,
   StyleSheet,
-  ViewPropTypes,
   requireNativeComponent,
   NativeModules,
   processColor,
 } from 'react-native';
+import { ViewPropTypes } from 'deprecated-react-native-prop-types';
```
luissantana, line 13: @emilioicai seems this new import is breaking the latest release as the dependency lib "deprecated-react-native-prop-types" is not included in package.json

emilioicai: sorry for that, v5.1.1 was just released with a fix for that

PR #877, opened 2022-05-18
Update RN to 0.68.2 + update dev dependencies

AlexanderEggers:
This PR acts as a base for myself to get started with the new architecture work. Main focus of this PR was to upgrade the project to support RN 0.68.2.

**This PR includes the following changes:**
- Using the RN 0.68.2 [upgrade helper ](https://react-native-community.github.io/upgrade-helper/) to make changes to the whole project.
- Update dev dependencies - there were plenty which were extremely outdated and I have not noticed any issues so far. Please correct me if I missed something. I am happy to revert some of these dependency updates.
- Update npm offline packages. That seems like a lot of changes. Do you think it would make more sense to disable that feature? Is there actually a case when you use these offline caches?
- Add ability to inject the lottie-android version via the consumer base project (via "reactNativeAndroidLottieVersion"). Something which I always thought is strange that this is missing. I can move that into a separate PR but I thought to sneak this into the current PR. 😆 
- Allow ndkVersion to be injected from the consumer base project. That is needed because M1/non-M1 machine are needing to use different ndk versions.

**Testing**
- iOS is tested via the example project which is still running without any problems.
- @emilioicai I will need your help with Android. I have tried to get that running for quite a while but I just cannot get the packager to run correctly. Because of that the example app cannot load the relevant bundle. I am sure it is something small I am missing.

**Next**
I would start adding the actual RN new architecture code as part of the next PR. I think I should find some time in the next weekends for that.

**Note for the upcoming new architecture work**
@emilioicai I am planning to build a new fabric example project which is just showing a single lottie animation for the MVP. Unfortunately we have to wait for the community slider and picker to support fabric to just re-usage the existing example project.


AlexanderEggers:
@emilioicai When starting to work on Fabric I noticed that the current project setup does not work well for that. I decided to attempt to fully refactor the structure - I hope you don't mind. I will put the changes of this PR into that same PR which I will open in a short while.

PR #901, opened 2022-07-12
Update readme to use new lottie-ios version

AlexanderEggers:
This PR should probably only be merged after doing the next release which actually unblocks the usage of the newer lottie-ios version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,16 +10,16 @@ For the first time, designers can create **and ship** beautiful animations witho
 ## Installing
 
 ### iOS and Android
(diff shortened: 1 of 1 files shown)


emilioicai (approved):

walterholohan:
@emilioicai will you be doing a new release, as currently, 5.1.3 does not support lottie-ios 3.4.0, but your readme says to install that version

emilioicai:
Yes, I'm preparing 6.0.0 so it's taking a bit longer

matinzd:
How long does releasing this going to take? You have already updated the document version but v6.0.0 is not out.

PR #907, opened 2022-07-27
Implement render mode property for iOS

jhen0409:
- Upgrade lottie-ios to v3.4.1, which included fixes such as https://github.com/airbnb/lottie-ios/pull/1660
- Implement render mode for iOS, so we can switch render engine by set `renderMode` prop
  - AUTOMATIC: Select Core Animation and fallback to Main Thread if not supported
  - HARDWARE: Core Animation ([More info](https://github.com/airbnb/lottie-ios/discussions/1627))
  - SOFTWARE: Main Thread
- Update `renderMode` in API documentation
- Example: Add render mode picker


jhen0409:
Resolved the conflicts.

For example, I also changed the render mode picker to use a dropdown, which reduces the height used.

matinzd (approved):
Looks good to me! Thanks for your contribution.

PR #915, opened 2022-08-12
[fix:iOS]Updated play method for animation with remote source url

Deepak-HDFC:
Instead of calling play on animation view using instance play method will add the animation finish event block.


emilioicai:
Have you tried running the local example? It fails for me when source is not remote.

PR #921, opened 2022-08-13
chore: remove lottie-ios peer dependency

matinzd:
## Problem
Users need to install `lottie-ios` compatible with their `lottie-react-native` version in order to use library.

## Action
As we are moving forward, I think it is good to remove peer dependency and depend on podspec to install the proper version from podspec without manually installing `lottie-ios` on the project.  
From now on, `lottie-ios` can be installed and detected automatically from podspec configuration and users do not need to install both of them in order to use it. 

## Checklist
- [x] Test iOS build

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,16 +10,16 @@ For the first time, designers can create **and ship** beautiful animations witho
 ## Installing
 
 ### iOS and Android
-Install `lottie-react-native` (latest) and `lottie-ios` (3.4.0):
+Install `lottie-react-native` (latest):
 
 ```
-yarn add lottie-react-native lottie-ios@3.4.0
+yarn add lottie-react-native
 ```
 
 or
 
 ```
-npm i --save lottie-react-native lottie-ios@3.4.0
+npm i --save lottie-react-native
 ```
 
 Go to your ios folder and run:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -10,16 +10,16 @@ For the first time, designers can create **and ship** beautiful animations witho
 ## Installing
 
 ### iOS and Android
-Install `lottie-react-native` (latest) and `lottie-ios` (3.4.0):
+Install `lottie-react-native` (latest):
 
 ```
-yarn add lottie-react-native lottie-ios@3.4.0
```
emilioicai, line 16: This may lead to future issues as new versions of `lottie-ios` are released. Having it as a peer dependency creates certainty of what's released works as expected.

matinzd: But we are forcing it in our podspec and it will work with our compatible version.

emilioicai: That's fine for new installations but upgrades will end up with lots of versioning issues.

emilioicai: Wait, since we are now aiming for a major version release we might include this one safely and add a migration document.

emilioicai (approved):

README.md
matinzd: Yes, we need to include it in v6.0.0 document.

PR #955, opened 2022-12-16
Fabric support for iOS

alfonsocj:
# Summary

Adding Fabric support for iOS.

- Update podspec file to match new architecture's paved path.
- Implementation of `LottieAnimationViewComponentView` objective-c++ class that uses existing `ContainerView.swift`.


packages/core/ios/Fabric/LottieAnimationViewComponentView.h
```diff
@@ -0,0 +1,26 @@
+#ifdef __cplusplus
```
alfonsocj, line 1: prevents issue `lottie-react-native/apps/fabric/ios/Pods/Headers/Public/React-Fabric/react/renderer/core/EventQueue.h:10:10: 'memory' file not found`

packages/core/ios/Fabric/LottieAnimationViewComponentView.mm
```diff
@@ -0,0 +1,125 @@
...
+        [_view setSpeed:newLottieProps.speed];
+    }
+    
+    if(oldLottieProps.colorFilters != newLottieProps.colorFilters) {
+        [_view setColorFilters:convertColorFilters(newLottieProps.colorFilters)];
+    }
+    
+    if(oldLottieProps.textFiltersIOS != newLottieProps.textFiltersIOS) {
+        [_view setTextFiltersIOS:convertTextFilters(newLottieProps.textFiltersIOS)];
+    }
```
alfonsocj, line 80: These comparisons work because of the struct operators set [above](https://github.com/lottie-react-native/lottie-react-native/pull/955/files#diff-7e1b40b8d45473fba9a66e1693cee215752e9f2d8f845d6dfd13059021529deaR13-R20)

packages/core/ios/Fabric/LottieContainerView.h
```diff
@@ -0,0 +1,28 @@
+@class UITraitCollection;
+@class NSDictionary;
+@class NSString;
+@class NSCoder;
+
+#import <React/RCTView.h>
+
+@interface LottieContainerView : RCTView
```
alfonsocj, line 8: Even though this swift class is available in objective-c because we set the `@objc(LottieContainerView)`, we have to create this interface manually because the class is `internal`.

packages/core/ios/Fabric/RNLottieFabricConversions.h
```diff
@@ -0,0 +1,29 @@
+#ifdef __cplusplus
+#import <React/RCTConversions.h>
+
+inline NSArray<NSDictionary *> *convertColorFilters(std::vector<facebook::react::LottieAnimationViewColorFiltersStruct> colorFilterStructArr)
```
alfonsocj, line 4: Very similar to the old arch `LRNColorFilters` `RCTConvert` implementation

packages/core/ios/LottieReactNative/ContainerView.swift
```diff
@@ -234,7 +242,9 @@ class ContainerView: RCTView {
         animationView = next
         addSubview(next)
         animationView?.contentMode = contentMode
-        animationView?.reactSetFrame(frame)
+        print("self frame \(frame)")
+        animationView?.reactSetFrame(UIScreen.main.bounds)
```
alfonsocj, line 246: TODO: Figure out how to get the right frame here

alfonsocj:
@AlexanderEggers, here's what I have so far. The fabric example works, but I still have to figure out a `frame` issue that prevents ContainerView's animationView not visible (because the frame is `(0.0, 0.0, 0.0, 0.0)`) [here](https://github.com/lottie-react-native/lottie-react-native/pull/955/files#diff-4990c7973a2de25dc8c6b0da5cee955784894050cdc7b031ed22d743ce9492f9R246)

packages/core/src/specs/LottieAnimationViewNativeComponent.ts
```diff
@@ -27,9 +37,11 @@ export interface NativeProps extends ViewProps {
   enableMergePathsAndroidForKitKatAndAbove?: boolean;
   hardwareAccelerationAndroid?: boolean;
   cacheComposition?: boolean;
-  colorFilters?: ReadonlyArray<string>;
+  colorFilters?: ReadonlyArray<ColorFilterStruct>;
```
alfonsocj, line 40: I just realized that `LottieView.tsx` converts `colorFilters` and `textFilters` into `string[]`. I wonder if we should get rid of the extra serialization

AlexanderEggers: I gave that a try but it didn't work for me. It might be solved with a newer RN version but the last time I checked the Android build failed to generate non-broken code via that type spec definition.

packages/core/ios/Fabric/LottieAnimationViewComponentView.h
AlexanderEggers: Did you find that in any specific documentation? Just in case there were more information around that specific line - I would be keen to learn a little more about that 😃

alfonsocj: No documentation. I had a similar issue a few months ago with Expo modules on the new arch and the Expo team helped me figure this out.

alfonsocj:
@AlexanderEggers I tested on Android and it seems like it's accepting the structs. The problem I see now is I get this error when the animation ends 
```java.lang.AssertionError: Method overloading is unsupported: com.facebook.react.uimanager.events.RCTModernEventEmitter#receiveEvent```
This happens when the `sendOnAnimationFinishEvent` method runs and `receiveEvent` gets called. Can you reproduce this? It's happening to me on the fabric example in master as long as `loop` prop is set to false.

**Comment by alfonsocj** [comment]
I believe that we are going to have to get the frame's size some other way. I had to do something similar in here https://github.com/react-native-datetimepicker/datetimepicker/blob/master/ios/fabric/cpp/react/renderer/components/RNDateTimePicker/RNDateTimePickerState.h#L19.

File: packages/core/ios/LottieReactNative/ContainerView.swift [hunk 7]
```diff
@@ -190,43 +197,58 @@ class ContainerView: RCTView {
...
         animationView?.pause()
     }
 
-    func resume() {
+    @objc func resume() {
         play()
     }
 
+    // reset the animationView frame whenever the view's frame changes
+    override var frame: CGRect {
```
alfonsocj, line 245: By the time we set the animationView's frame, React might not have calculated the view's frame so we need to change the animationView's frame when the view's frame changes.

alfonsocj:
@emilioicai @AlexanderEggers @matinzd I think the PR is ready!

matinzd:
I am going to test it this weekend. Thanks for the PR!

packages/core/ios/LottieReactNative/ContainerView.swift
```diff
@@ -11,9 +16,11 @@ class ContainerView: RCTView {
     private var colorFilters: [NSDictionary] = []
     private var textFilters: [NSDictionary] = []
     private var renderMode: RenderingEngineOption = .automatic
+    @objc weak var delegate: LottieContainerViewDelegate? = nil
```
alfonsocj, line 19: New delegate so that `LottieAnimationViewComponentView` can send the `animationFinish` event to JS

packages/core/ios/Fabric/LottieAnimationViewComponentView.h
```diff
@@ -0,0 +1,27 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface LottieAnimationViewComponentView : RCTViewComponentView <LottieContainerViewDelegate>
+@end
+
+namespace facebook {
+namespace react {
+// In order to compare these structs we need to add the == operator for each
+bool operator==(const LottieAnimationViewColorFiltersStruct& a, const LottieAnimationViewColorFiltersStruct& b)
```
alfonsocj, line 14: https://github.com/reactwg/react-native-new-architecture/discussions/91#discussioncomment-4426469

packages/core/src/specs/LottieAnimationViewNativeComponent.ts
```diff
@@ -27,9 +42,11 @@ export interface NativeProps extends ViewProps {
   enableMergePathsAndroidForKitKatAndAbove?: boolean;
   hardwareAccelerationAndroid?: boolean;
   cacheComposition?: boolean;
-  colorFilters?: ReadonlyArray<string>;
-  textFiltersAndroid?: ReadonlyArray<string>;
-  textFiltersIOS?: ReadonlyArray<string>;
+  colorFilters?: ReadonlyArray<ColorFilterStruct>;
+  // dummy that solves codegen issue when there's a ReadonlyArray<Object> without another Object prop
+  dummy?: Readonly<{ dummy: boolean }>;
```
AlexanderEggers, line 47: Is there an existing issue for that problem? That seems like a very strange way to solve this codegen problem.

packages/core/lottie-react-native.podspec
```diff
@@ -19,24 +19,26 @@ Pod::Spec.new do |s|
...
-    "CLANG_CXX_LANGUAGE_STANDARD" => "c++17",
-  }
-
   if ENV['RCT_NEW_ARCH_ENABLED'] == '1' then
-    s.subspec "fabric" do |ss|
-      ss.dependency "React-RCTFabric"
-      ss.dependency "React-Codegen"
-      ss.source_files       = "ios/Fabric/**/*.{h,m,mm,swift}"
-      ss.exclude_files      = "ios/**/*.{swift}"
-    end
```
AlexanderEggers, line 40: I am wondering why you removed that part. I actually added that so the fabric files would be provided via a subspec and would have a separate entry in the podfile.lock.

alfonsocj: Hey, I can bring it back. I preferred to have this as similar as possible to the one Meta has been recommending. It also keeps it consistent with other libraries.

packages/core/src/specs/LottieAnimationViewNativeComponent.ts
alfonsocj: I'm creating it to add more visibility. I had a conversation about it with cipolleschi and Titozzz on discord. https://github.com/reactwg/react-native-new-architecture/discussions/104

alfonsocj:
Happy New Year! @matinzd, did you have a chance to test it?

matinzd:
> Happy New Year! @matinzd, did you have a chance to test it?

I am going to test this tonight. 🚀

matinzd:
I ran into some issues while building and testing it last night. I will give more update on it later today.

packages/core/lottie-react-native.podspec
matinzd: Let's keep it as recommended and it would help ensure that the library is consistent with other libraries and that any potential bugs are more likely to be discovered and fixed more quickly.

matinzd:
Try playing with examples and adding the view to different hierarchies. It starts throwing an error and it renders nothing after that.

Screenshot 2023-01-05 at 09 22 15

bukhari-flip:
I think the `Cannot read property '_nativeTag' of null` appeared when fast reload was enabled. Because when an error appeared and then you force to refresh the app, the error will disappear.

This behavior happened on both iOS and Android.

matinzd:
> I think the `Cannot read property '_nativeTag' of null` appeared when fast reload was enabled. Because when an error appeared and then you force to refresh the app, the error will disappear.
> 
> 
> 
> This behavior happened on both iOS and Android.

I think we might need to refactor javascript code as well. There are some deprecated methods used in our js files.

alfonsocj:
@matinzd, which deprecated methods are you referring to? I'd be happy to refactor those.

matinzd:
> @matinzd, which deprecated methods are you referring to? I'd be happy to refactor those. Are you referring to `AnimatedLottieView` being a class based component with lifecycle events like `componentDidUpdate`?

Sorry, my bad. I thought we are still using `dispatchViewManagerCommand` and `dispatchCommand`.

packages/core/src/specs/LottieAnimationViewNativeComponent.ts
matinzd: But on android codegen will just generate `ReadableArray`, right? @alfonsocj

alfonsocj:
I'm not sure why that's happening on Fabric when using fast refresh. It's related to the native component having a ref. I couldn't reproduce the issue when removing the ref from the native component in `LottieView.tsx`.

apps/fabric/ios/Podfile.lock
```diff
@@ -16,13 +16,12 @@ PODS:
   - lottie-ios (3.5.0)
   - lottie-react-native (6.0.0-rc.1):
     - lottie-ios (~> 3.5.0)
-    - lottie-react-native/fabric (= 6.0.0-rc.1)
-    - React-Core
-  - lottie-react-native/fabric (6.0.0-rc.1):
-    - lottie-ios (~> 3.5.0)
+    - RCT-Folly
```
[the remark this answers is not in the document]
matinzd: Is this done? If so, let's resolve this.

packages/core/ios/LottieReactNative/ContainerView.swift
matinzd: Cool! That was broken for some time I AFAIK.

alfonsocj:
@matinzd, let me know when you can test my findings. I'll investigate further what's going on with the refs.

packages/core/src/specs/LottieAnimationViewNativeComponent.ts
matinzd: Is this solved? I am still dealing with it in react native 0.71-rc5. Can you verify?

packages/core/ios/Fabric/LottieContainerView.h
matinzd: Can I mark this as resolved?
Or maybe add your comment in the code so that we can refer back to it.

packages/core/ios/Fabric/LottieAnimationViewComponentView.h
matinzd: Please add this link as a reference inside the code and mark it as TODO or FIX.

packages/core/ios/Fabric/RNLottieFabricConversions.h
matinzd: Can we add some color filters in the example app to make sure this works properly?

matinzd:
> @matinzd, let me know when you can test my findings. I'll investigate further what's going on with the refs.

Ok, I found the issue. According to this [note](https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs) first callback of ref will be initiated as `null` during the first render of DOM (it is related to react internals). So if we wrap ref with null check it fixes the problem.

```diff
  _captureRef(ref: React.ElementRef<typeof NativeLottieAnimationView>) {
+   if (ref !== null) {
+      this._lottieAnimationViewRef = ref;
+    }
    if (this.props.autoPlay === true) {
      this.play();
    }
  }
```

packages/core/ios/LottieReactNative/ContainerView.swift
alfonsocj: hmm this will only affect the new arch. I tested it on the current arch and it works fine.

packages/core/src/specs/LottieAnimationViewNativeComponent.ts
alfonsocj: It hasn't been solved. I'll link the discussion here.

alfonsocj:
I added color filters to the example. I see those working fine on iOS. Android is not getting them.

alfonsocj:
Android is getting the values and assigning a color filter. I think the problem is that the color filter keypath for Android is not the same than for iOS.

alfonsocj:
@matinzd have you tried using a lottie animation that includes a text before? I tried one that uses Roboto font and I couldn't get it to work on Android since the font is not there. I tried adding it to `assets/font` but that didn't work.

matinzd:
Hey @alfonsocj ,
Example upgrade is ready! Try rebasing/updating your branch.

matinzd (approved):
Looks good! 🚀

PR #956, opened 2022-12-20
Update README.md for more consistency

skantus:
For the version of `lottie-react-native 5.1.4` no need to be installed separately since `lottie-ios` is installed automatically to `v3.4.4`, and then there is no limitation to have only the main dependency `lottie-react-native`.

Screenshot 2022-12-20 at 21 24 27

Screenshot 2022-12-20 at 20 56 36

Tested on iOS devices.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Install `lottie-react-native` (latest):
 yarn add lottie-react-native
 ```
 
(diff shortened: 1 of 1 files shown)


matinzd:
This was mostly for backward compatibility since we did not test those versions with the current wrapper. I think it's a better idea to keep it as it is to decrease confusion in the community.

PR #962, opened 2023-01-11
chore: upgrade examples to 0.71.0

matinzd:
Upgrade all examples to `0.71.0` to prepare for rolling out new architecture.


yarn.lock
```diff
@@ -85,7 +85,7 @@ __metadata:
   languageName: node
```
emilioicai, line 85: do we really need to update the root `yarn.lock`?

matinzd: Yes. We are using yarn workspaces and the only `yarn.lock` we have is in root of the project.

emilioicai (approved):

matinzd:
@AlexanderEggers Can you check the Android paper example here and see if you can build it?

AlexanderEggers:
Sure I will have a look this weekend 👍

AlexanderEggers:
@matinzd I had a look at the changes + did a few local changes that I can push up if you want? I am actually unable to build the paper apps using the RN 0.71 upgrade. It seems to be related to certain sub-dependencies not supporting RN 0.71 yet which throws error like the following for me:

```
* What went wrong:
Could not determine the dependencies of task ':app:compileDebugJavaWithJavac'.
> Could not resolve all task dependencies for configuration ':app:debugCompileClasspath'.
   > Could not find com.facebook.react:react-android:.
     Required by:
         project :app
   > Could not find com.facebook.react:hermes-android:.
     Required by:
         project :app
   > Could not find any matches for com.facebook.react:react-native:+ as no versions of com.facebook.react:react-native are available.
     Required by:
         project :app > project :react-native-community_slider
         project :app > project :react-native-picker_picker
```

We could try to patch these sub-dependencies but I am not certain if that is the suggested way by the RN team.

cc @emilioicai

matinzd:
@AlexanderEggers I had the same issue. I think we can fix it by overriding sub dependecies with gradle on top or for now you can just patch them and then we can come up with a better solution.

AlexanderEggers:
Ok, I will take care of patching then 👍 Should I raise a separate PR for that work or do I have the right repo permissions to push directly into this branch?

It is probably worth raising this with RN if that wasn't discussed somewhere before.

matinzd:
> Ok, I will take care of patching then 👍 Should I raise a separate PR for that work or do I have the right repo permissions to push directly into this branch?
> 
> It is probably worth raising this with RN if that wasn't discussed somewhere before.

You can make a PR based on this branch.

AlexanderEggers:
@matinzd I tried patching these libs but no luck to resolve all errors. For some reasons the react plugin is not properly resolved.

matinzd:
I will also take a look into that when I find a free time. Thank you! @AlexanderEggers

matinzd:
@AlexanderEggers I figured out that `react-native-windows` is causing conflicts. 
I submitted an issue [here](https://github.com/microsoft/react-native-windows/issues/11093).

AlexanderEggers:
@matinzd Should we remove react-native-windows from the paper-example project then?

matinzd:
> @matinzd Should we remove react-native-windows from the paper-example project then?

I don't know. I was thinking maybe dropping support for windows for now if possible. But @emilioicai should comment on that. 
But unfortunately we are using it and we are supporting windows.

emilioicai:
iOS and Android support are more important now. We could re-add the support as a separate initiative and recommend the current version which seems to be working fine for current windows users.

PR #975, opened 2023-01-26
fix: add missing import and update path in readme

tomekzaw:
This PR introduces two minor fixes to README:
* Add missing `useRef` import in one code example
* Change example path in another example for consistency


matinzd (approved):
Thanks for your contribution!

Issue #825, opened 2021-12-08
animation reset to first frame when app resume from background


sbin1211:
### Description

at ios, after animation finished, when app goto background then resume, the animation reset to first frame not end of frame. 
this issue accure at 5.0.0, 5.0.1, it's ok before 5.0.0 

### Steps to Reproduce

https://user-images.githubusercontent.com/22540308/145149332-3ff9113d-c39a-4d73-8f58-58ea9c779c83.mp4

**Expected behavior:** 
animation should stay at last of frame.

**Actual behavior:** 
animation reset to first frame.

### Versions
lottie-react-native@5.0.1
react-native@0.66
system

sbin1211:
@emilioicai could you have a look. thanks.

nathanlevitt:
This is happening to mine as well, along with looped animations causing the frame to freeze when app resumes from background.

kieranoreilly:
Having same issue :')

kyang6:
Same here, any insights?

kyang6:
https://github.com/lottie-react-native/lottie-react-native/issues/412 I think related to this issue from a while back

Livijn:
I have the same problem. My workaround looks like this:

```js
  useEffect(() => {
    AppState.addEventListener('change', handleAppStateChange);

    return () => {
      AppState.removeEventListener('change', handleAppStateChange);
    };
  }, []);

  function handleAppStateChange(nextAppState: AppStateStatus) {
    if (nextAppState === 'active') {
      lottieRef.current!.play(75, 75);
    }
  }
```

cam-shaw:
Happening for us on iOS, 5.0.1

branoholy:
I had the same issue and the workaround from @Livijn didn't solve it completely. The animation blinked from the first frame to the last frame. Specifying the property `progress={1}` helped in my case. Fortunately, `autoPlay` still works as expected. I use these properties.

```tsx
<LottieView
  source={require('...')}
  autoPlay
  progress={1}
  loop={false}
/>
```

Livijn:
@branoholy Actually, it seems that setting `progress={1}` worked. 😮

MiteshKalal7:
@Livijn  Thanks
```  
const lottieRef = useRef();

useEffect(() => {
    const sub = AppState.addEventListener("change", handleAppStateChange);

    return () => sub.remove();
  }, []);

  const handleAppStateChange = (nextAppState) => {
[...]

PR #980, opened 2023-01-31
fix(iOS): reset to first frame on app resume bug

alfonsocj:
This fixes the bug where the animation resets to its first frame when going to background and resume.
Issue: https://github.com/lottie-react-native/lottie-react-native/issues/825

## Why this problem occured?
When going to background, `traitCollectionDidChange` method runs which calls `applyProperties()`. This changes the animation's current progress `animationView.currentProgress = progress` and `progress` comes from props. If not defined it defaults to 0 (first frame).

Additionally, setting `backgroundBehavior = .pauseAndRestore` has a weird behavior. When the animation ends and the app goes to background, it sets the animation to the first frame. Forcing the `currentProgress` to the last frame when the animation ends works fine to mitigate the issue.

## Before 
https://user-images.githubusercontent.com/3791120/215833991-ec57dd77-9f7e-4802-b846-0df7f9a42cbf.mov

## After
https://user-images.githubusercontent.com/3791120/215833635-af090c2f-e6cb-4367-b152-bce21b1419b7.mov

### DIFF
--- a/packages/core/ios/LottieReactNative/ContainerView.swift
+++ b/packages/core/ios/LottieReactNative/ContainerView.swift
@@ -25,7 +25,7 @@ class ContainerView: RCTView {
         super.traitCollectionDidChange(previousTraitCollection)
         if #available(iOS 13.0, tvOS 13.0, *) {
             if (self.traitCollection.hasDifferentColorAppearance(comparedTo: previousTraitCollection)) {
-                applyProperties()
+                applyColorProperties()
                 print("dark mode changed")
             }
         }
@@ -194,7 +194,7 @@ class ContainerView: RCTView {
 
     @objc func setColorFilters(_ newColorFilters: [NSDictionary]) {
         colorFilters = newColorFilters
-        applyProperties()
+        applyColorProperties()
     }
 
     // There is no Nullable CGFloat in Objective-C, so this function uses a Nullable NSNumber and converts it later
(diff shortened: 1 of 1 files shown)


packages/core/ios/LottieReactNative/ContainerView.swift
```diff
@@ -257,15 +261,15 @@ class ContainerView: RCTView {
...
-        applyProperties()
+        animationView?.backgroundBehavior = .pauseAndRestore
+        animationView?.animationSpeed = speed
+        animationView?.loopMode = loop
+        applyColorProperties()
     }
-
-    func applyProperties() {
+    
+    func applyColorProperties() {
```
alfonsocj, line 270: Only apply color properties here. Setting animationSpeed or loopMode resets the animation

packages/core/ios/LottieReactNative/ContainerView.swift
```diff
@@ -210,9 +210,11 @@ class ContainerView: RCTView {
                 onFinish(["isCancelled": !animationFinished])
             }
             self.delegate?.onAnimationFinish(isCancelled: !animationFinished);
+            if (animationFinished) {
+                self.animationView?.currentProgress = 1;
```
alfonsocj, line 214: background behavior `.pauseAndRestore` works fine until the animation ends. If the animation has ended it goes back to the first frame. By setting the progress to the last frame once it finishes, it forces the animation to stay on the last frame.

alfonsocj:
@matinzd I think it's ready for review.

matinzd (approved):

PR #991, opened 2023-02-21
docs: update Lottie version

MateWW:
Hey,
I just noticed that we have the wrong version in `README.md`

Thanks for delivering such a good piece of code :)


matinzd:
Hey,

Thanks for your contribution. You don't need to install `lottie-ios` separately anymore since 5.1.5.

PR #994, opened 2023-03-03
Update Doc

vijaychouhan-rails:
Doc Updated: Change support version 5.1.4 to 5.1.5


matinzd:
Thanks for your suggestion. You don't need to install `lottie-ios` separately and the readme is there intentionally.

Issue #963, opened 2023-01-12
[Android] Crash on new arch when lottie animation finishes


alfonsocj:
### Description
Lottie animation crashes the app when the animation finish event is triggered.
Only reproducible when the RN new arch is enabled.

### Steps to Reproduce

- Open `apps/fabric/App.tsx`
- Set `loop={false}` (or remove `loop`) in `AnimatedLottieView`
- Run `yarn android` in `apps/fabric`
- Wait for the animation to finish.

**Expected behavior:** [What you expected to happen]
The animation finishes without the app crashing

**Actual behavior:** [What actually happened]
App crashes with:
```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.airbnb.android.lottie.rnexample, PID: 6274
    java.lang.AssertionError: Method overloading is unsupported: com.facebook.react.uimanager.events.RCTModernEventEmitter#receiveEvent
        at com.facebook.react.bridge.JavaScriptModuleRegistry$JavaScriptModuleInvocationHandler.<init>(JavaScriptModuleRegistry.java:66)
        at com.facebook.react.bridge.JavaScriptModuleRegistry.getJavaScriptModule(JavaScriptModuleRegistry.java:41)
        at com.facebook.react.bridge.CatalystInstanceImpl.getJSModule(CatalystInstanceImpl.java:464)
        at com.facebook.react.bridge.ReactContext.getJSModule(ReactContext.java:160)
        at com.airbnb.android.react.lottie.LottieAnimationViewManager.sendOnAnimationFinishEvent(LottieAnimationViewManager.kt:57)
        at com.airbnb.android.react.lottie.LottieAnimationViewManager.access$sendOnAnimationFinishEvent(LottieAnimationViewManager.kt:31)
        at com.airbnb.android.react.lottie.LottieAnimationViewManager$createViewInstance$1.onAnimationEnd(LottieAnimationViewManager.kt:87)
        at android.animation.Animator$AnimatorListener.onAnimationEnd(Animator.java:600)
        at com.airbnb.lottie.utils.BaseLottieAnimator.notifyEnd(BaseLottieAnimator.java:74)
        at com.airbnb.lottie.utils.LottieValueAnimator.doFrame(LottieValueAnimator.java:103)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1229)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1239)
[...]

PR #1000, opened 2023-03-26
fix: refactor event dispatch logic on android to fix fabric crash

AlexanderEggers:
fixes https://github.com/lottie-react-native/lottie-react-native/issues/963

This PR refactors the event dispatch logic on Android and makes use of the EventDispatcher on paper and fabric. I did a test in both architectures via the example project and was successfully able to see the relevant console log which is running when an animation finishes.

cc @matinzd @alfonsocj @emilioicai


matinzd (approved):
Thank you a lot for helping me out with this. Unfortunately, I couldn't find free time to fix this.
I tested it a lot and it looks great!

PR #1019, opened 2023-05-03
Fix iOS frames being calculated incorrectly

TheRogue76:
As of RC4 ([This PR](https://github.com/lottie-react-native/lottie-react-native/pull/992)), Lottie no longer wraps the Native component in a separate view, and passes all the view props to the native implementation. This causes an issue with Lottie on iOS, as the iOS version of Lottie is made of two different views:

1. A `RCTView` from react native that gets all of the props (Like background color, etc)
2. A `LottieAnimationView` which is a `Subview` of the `RCTView`

The `LottieAnimationView` is tapping into the `RCTView`'s `frame` and giving it to itself. But this is incorrect as the `frame` property on iOS [has the coordinates of the `RCTView` as well](https://suragch.medium.com/frame-vs-bounds-in-ios-107990ad53ee#:~:text=Since%20frame%20relates%20a%20view's,arranging%20subviews%20within%20the%20view.), leading to a rather gnarly offset. Instead of `frame`, the `bounds` property should be used.

This PR does three things:

1. Update both example apps to 71.7 to fix the issue of the newest version of XCode not building
2. Changes the property wrapper for setting the frame of `LottieAnimationView` from `frame` to `bounds` and removes the hacky override at the top. 
3. Document how the setup for the views on iOS works at the top of the file

*NOTE*: The reason wrapping the Native component with a view fixed the issue is that the `frame` of the `RCTView` would be equal to the `bounds`, due to the styling being applied to the parent and the fact that `frame` is calculated *relative* to the parent


matinzd (approved):
Looks good! Thanks for your contribution.

PR #1021, opened 2023-05-04
Fix autoplay for React Navigation

elliottkember:
I've found that sometimes, in React Native, autoPlay doesn't work. This seems to be a timing issue, where the element isn't already on the screen. 

Waiting for one frame before playing seems to solve the issue reliably for me. I've been doing this manually in components that use Lottie animations, so I figured I would submit this PR to fix it upstream.

I don't know of any issues that could be caused by adding this timeout around this play call, but open to suggestions.


matinzd:
Is this just happening in React Navigation? Are you using freeze feature?

matinzd:
Closing due to inactivity.

elliottkember:
Whoops - I didn't see your earlier comment. Yes it was happening in RN and we were not using "freeze". It is a native-stack though so using RNScreens which may affect things.

Issue #1018, opened 2023-05-01
Build crashing on React native macos


HunainHumail:
### Description
I have installed react-native-macos in my project. Android and IOS are working fine with react-native-lotte. But when i am trying to build the macos project, the build is crashing.

[Description of the bug]
The following build commands failed:
	SwiftEmitModule normal arm64 Emitting\ module\ for\ lottie_react_native (in target 'lottie-react-native-macOS' from project 'Pods')
(1 failure)

Also when I see my XCODE, getting this error Cannot find type 'UITraitCollection' in scope
### Steps to Reproduce

react-native run-macos

**Expected behavior:** [What you expected to happen]
App should work as it is working for android and ios
**Actual behavior:** [What actually happened]
App is crashing
<img width="1181" alt="Screenshot 2023-05-01 at 3 14 17 PM" src="https://user-images.githubusercontent.com/37023744/235439478-0860f6e3-3e66-4ee6-a0da-0e1c583dfa6e.png">

### Versions
react native: 0.71
react native macos: 0.71

You can get this information from executing `npm version`.

matinzd:
What version of lottie are you using?

HunainHumail:
> What version of lottie are you using?

"lottie-react-native": "^5.1.5",
"react": "18.2.0",
    "react-native": "0.71.7",
    "react-native-macos": "^0.71.0-0",

HunainHumail:
> What version of lottie are you using?

"lottie-react-native": "^5.1.5",
"react": "18.2.0",
"react-native": "0.71.7",
"react-native-macos": "^0.71.0-0",

Saadnajmi:
React Native macOS maintainer here, if we ifdef out the iOS block for macOS, are there any other issues? I took a quick look at the code and it seems to be implemented cross platform

matinzd:
Let me know if this has fixed the issue in rc7. @HunainHumail

PR #1031, opened 2023-05-25
fix: react native macos build

matinzd:
Fixes: https://github.com/lottie-react-native/lottie-react-native/issues/1018

### DIFF
--- a/packages/core/ios/LottieReactNative/ContainerView.swift
+++ b/packages/core/ios/LottieReactNative/ContainerView.swift
@@ -26,6 +26,7 @@ class ContainerView: RCTView {
     @objc var onAnimationFinish: RCTBubblingEventBlock?
     var animationView: LottieAnimationView?
     
+    #if !(os(OSX))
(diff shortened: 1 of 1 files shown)


Saadnajmi:
> /cc @Saadnajmi

Do we have a way to test macOS? What we've done in other repos is move the test app to use this (https://github.com/microsoft/react-native-test-app). That way, it is easy to add out of tree platforms like macOS and windows without needing to worry about maintaining extra boilerplate code.

matinzd:
> > /cc @Saadnajmi
> 
> Do we have a way to test macOS? What we've done in other repos is move the test app to use this (https://github.com/microsoft/react-native-test-app). That way, it is easy to add out of tree platforms like macOS and windows without needing to worry about maintaining extra boilerplate code.

Yeah this is what I was trying to go for but we somehow decided to not go with it that time.
Maybe it's a good time to move to RNTA.

emilioicai (approved):

Saadnajmi:
Awesome! :D

PR #1037, opened 2023-06-18
chore: update project dependencies

AlexanderEggers:
This PR updates different project dependencies and fixes a few inconsistencies between the workspaces in regards to which version should be used. For example the project was still enforcing react 18.1.0 or the correct metro babel plugin was not used in the core library workspace.


matinzd (approved):
Looks great! Thank you.

PR #1038, opened 2023-06-22
chore: add support for upcoming 0.73

matinzd:
Reference: https://github.com/react-native-community/discussions-and-proposals/issues/671

### DIFF
--- a/packages/core/android/build.gradle
+++ b/packages/core/android/build.gradle
@@ -33,6 +33,7 @@ if (isNewArchitectureEnabled()) {
 }
 
 android {
+    namespace = "com.airbnb.android.react.lottie"
     compileSdk getExtOrDefault('compileSdkVersion', 31)
 
     // Used to override the NDK path/version on internal CI or by allowing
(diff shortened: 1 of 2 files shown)


packages/core/android/build.gradle
```diff
@@ -33,6 +33,7 @@ if (isNewArchitectureEnabled()) {
 }
 
 android {
+    namespace = "com.airbnb.android.react.lottie"
```
AlexanderEggers, line 36: @matinzd What do you think of using something like instead? https://github.com/software-mansion/react-native-gesture-handler/blob/9be7bc38d529473f27ba03ca61444b0484c1f28f/android/build.gradle#L173 Just to avoid making the library incompatible for everyone else not using Gradle 7 yet.

matinzd: Yeah I was thinking the same? How about the manifest itself?

AlexanderEggers: I think that prop should be ignored in AGP 8 and we can just leave that value. That's hos other packages are solving that at the moment.

matinzd: Yeah makes sense. I also removed `INTERNET` permission from it.

AlexanderEggers (approved):
Changes look good to me 👍

emilioicai (approved):

PR #1045, opened 2023-07-10
Override `frame` in addition to `bounds`

j-piasecki:
In some cases (on the new architecture`) the frame of the `animationView` was not set correctly by the overridden `bounds`, while the `frame` was receiving the correct position & size.

I've added an override for `frame` which functions in the exact same way as the one for `bounds` and it solved the problem.

### DIFF
--- a/packages/core/ios/LottieReactNative/ContainerView.swift
+++ b/packages/core/ios/LottieReactNative/ContainerView.swift
@@ -251,6 +251,12 @@ class ContainerView: RCTView {
             animationView?.reactSetFrame(self.bounds)
         }
     }
+
+    override var frame: CGRect {
+        didSet {
+            animationView?.reactSetFrame(self.bounds)
+        }
+    }
     
     // MARK: Private
     func replaceAnimationView(next: LottieAnimationView) {


WoLewicki (approved):

matinzd:
Can you give me the reproduction of this issue? In which cases was this happening?

j-piasecki:
Sure, here you go: https://github.com/j-piasecki/lottie-rn-new-arch-repro. It's happening on the new architecture, but I'm not sure exactly what's causing it 😞.

When you install pods with `RCT_NEW_ARCH_ENABLED=1 pod install` there should be a blank screen, but after making the change from this PR the animation should appear.

TheRogue76:
From what i remember when fixing it in https://github.com/lottie-react-native/lottie-react-native/pull/1019, the frame property has the global position of the RCTView and adding it to the child view gives it an offset, strange that the new arch has a different behaviour. Maybe it would be best to test the behaviour in both modes, try to make sure nothing is broken? If they do have different behaviours we can investigate more

j-piasecki:
> the frame property has the global position of the RCTView and adding it to the child view gives it an offset

That's true, but I'm not using the `frame` anywhere, just the `didSet` observer on it to set the child's `frame` to the parent's `bounds`.



> Frame of animationView should be assigned to the bounds of the parent view.

Isn't that already there?
https://github.com/lottie-react-native/lottie-react-native/blob/0a8fff50a5a8d9ac12350b52f881ee015928f4af/packages/core/ios/LottieReactNative/ContainerView.swift#L244-L249
and
https://github.com/lottie-react-native/lottie-react-native/blob/0a8fff50a5a8d9ac12350b52f881ee015928f4af/packages/core/ios/LottieReactNative/ContainerView.swift#L259

matinzd:
Fixed in v6.0.1.

PR #1047, opened 2023-07-13
Docs/versioning

leoaraujocosta:
Description:

I was setting up this lib in one of my project and I realize the version in the README was outdated.

This PR is just and Docs update.

Hope that helps.


matinzd:
If you use the latest version you do not need to install `lottie-ios` separately anymore.

PR #1056, opened 2023-07-21
fix: memory leak issue

TheRogue76:
As per our discussion, This PR does a bunch of refactoring to both:

1. Simplify the logic of `setSourceURL` with a guard
2. Break the reference cycle (only needed for the internal Dispatch back to main because the first one doesn't capture self anyway)
3. Break the reference cycle for the callback

One area we can improve is with moving the call to decode Json in `setSourceJson` to a background thread, but it isn't a memory leak so much as "Kind of a slow op" so not in scope for this PR

I also made the completion callback a computed property so it doesn't recreate a new callback every time, just one, and while there was a small memory improvement, it was minuscule, so i am more than willing to roll it back

Picture of instruments showing a release:
Screenshot 2023-07-21 at 11 59 26


matinzd (approved):
Looks good!

PR #1057, opened 2023-07-22
feat: registration compat for 0.68/0.69

Sunbreak:
# Background

Support for Fabric on 0.68/0.69 requires manual registration inside custom `RNScreensComponentsRegistry`, according to [Introducing Fabric to react-native-screens](https://blog.swmansion.com/introducing-fabric-to-react-native-screens-fd17bf18858e), especially [Android base changes](https://github.com/software-mansion/react-native-screens/pull/1263)

# Solution

Backport `lottie-react-native` to 0.68/0.69 as https://github.com/callstack/react-native-builder-bob/pull/360

# Test Plan

Tested on `fabric-compat` example: https://github.com/Sunbreak/lottie-react-native/tree/compat-example

### DIFF
--- a/packages/core/android/Android.mk
+++ b/packages/core/android/Android.mk
@@ -1,0 +1,45 @@
+THIS_DIR := $(call my-dir)
+
+include $(REACT_ANDROID_DIR)/Android-prebuilt.mk
+
+include ${GENERATED_SRC_DIR}/codegen/jni/Android.mk
+
+include $(CLEAR_VARS)
+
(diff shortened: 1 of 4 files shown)


matinzd:
I don't think this is needed because I don't think it's a good idea to turn on new architecture on 68/69. New architecture support is still experimental and most production apps are not using it.

Sunbreak:
Fine. We could leave the PR open as a reminder

matinzd:
It's not needed anymore because we are not going to support versions lower than 70 anymore.

PR #1069, opened 2023-07-30
export types from lottie

ethanneff:
version 6.0.0 removed the types from lottie. this change should allow them to be imported again

```ts
src/apps/Focus/screens/Landing/index.tsx:1:23 - error TS2614: Module '"lottie-react-native"' has no exported member 'AnimationObject'. Did you mean to use 'import AnimationObject from "lottie-react-native"' instead?

1 import Lottie, { type AnimationObject } from 'lottie-react-native';
                        ~~~~~~~~~~~~~~~

```


matinzd (approved):
I updated examples in the repo as well.
Thank you!

PR #1072, opened 2023-07-31
Fix for tvOS

amochkin:
Fix for `pod install`/`pod update` error: 
`lottie-react-native-tvOS` does not specify a Swift version and none of the targets

### DIFF
--- a/packages/core/src/LottieView.tsx
+++ b/packages/core/src/LottieView.tsx
@@ -32,7 +32,7 @@ const defaultProps: Props = {
 export class LottieView extends React.PureComponent<Props, {}> {
   static defaultProps = defaultProps;
 
-  _lottieAnimationViewRef:
+  private lottieAnimationViewRef:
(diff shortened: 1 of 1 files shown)


TheRogue76:
Hi @amochkin 

Thank you for your contribution. Swift 5.0 is a bit old at this point and i don't believe we should be supporting that far back. 

@matinzd @emilioicai I think we can adopt a policy of N-2 for the swift version to support (considering they are bound to Xcode versions, we should not allow too many older versions). In this case, as the latest swift version is 5.8 (with 5.9 coming in a couple of months) we would be doing minimum support of 5.6, which is Xcode 13.3 (Link to version of Swift and Xcode: https://swiftversion.net)

Any thoughts?

amochkin:
Hi, @TheRogue76. I see no problem in declaring Swift version 5.8. Let me test that first and I will modify or make another PR.

PR #1071, opened 2023-07-30
feat: dotLottie support for iOS

matinzd:
## Rationale

Hey there! So, the community has been buzzing with requests for dotLottie file support, and we totally get why! 😄 People are super excited about enhancing the visual and interactive aspects of React Native apps, and dotLottie files can help us achieve just that!

The reason behind implementing this is even more strengthened by the effective integration of dotLottie support in the iOS and Android editions of Lottie. This is proven by the pull requests provided below:

iOS implementation: https://github.com/airbnb/lottie-ios/pull/1785
Android implementation: https://github.com/airbnb/lottie-android/pull/1660

## Related issues

To address existing concerns and ensure a comprehensive implementation, we acknowledge the following related issues:

- https://github.com/lottie-react-native/lottie-react-native/issues/1032
- https://github.com/lottie-react-native/lottie-react-native/issues/996
- https://github.com/lottie-react-native/lottie-react-native/issues/818
- https://github.com/Expensify/App/issues/15053

## Testing checklist

- [x] Previous implementations should work as expected (Backward compatibility)
- [x] Make sure DotLottie works both for remote and local assets
- [x] Update documentation for metro configuration
- [x] Monitor performance 
- [x] Test on older iOS devices

## Demo video

https://github.com/lottie-react-native/lottie-react-native/assets/24797481/b07cdf8a-3548-4c9c-8fbb-4f372e8f9a5a

PR #1074, opened 2023-08-02
feat: dotLottie support for android

matinzd:
Related to: https://github.com/lottie-react-native/lottie-react-native/pull/1071

### DIFF
--- a/packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManagerImpl.kt
+++ b/packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManagerImpl.kt
@@ -167,6 +167,15 @@ internal object LottieAnimationViewManagerImpl {
     }
 
     @JvmStatic
+    fun setSourceDotLottieURI(
+        uri: String?,
+        viewManager: LottieAnimationViewPropertyManager
+    ) {
+        viewManager.sourceDotLottie = uri
+        viewManager.commitChanges()
+    }
+
+    @JvmStatic
     fun setCacheComposition(view: LottieAnimationView, cacheComposition: Boolean) {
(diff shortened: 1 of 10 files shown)


packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManagerImpl.kt
```diff
@@ -286,4 +286,13 @@ internal object LottieAnimationViewManagerImpl {
     ) {
         viewManager.textFilters = textFilters
     }
+
+    @JvmStatic
+    fun setSourceDotLottieURI(
+        uri: String?,
+        viewManager: LottieAnimationViewPropertyManager
+    ) {
+        viewManager.sourceDotLottie = uri
```
TheRogue76, line 295: Considering what react bridge does when a field goes from string to undefined on iOS, do you think we should also add the clean up here as well for Android?

TheRogue76: Nit: Also please move this up under line 168 so all the source settings are in the same place

TheRogue76 (commented):
Looks Good. Just two comments and the only blocker is validating that the values get reset correctly when switching from source methods

packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManagerImpl.kt
matinzd: For Android we have null check everywhere `LottieAnimationViewPropertyManager`. I moving it up.

TheRogue76 (approved):

PR #1085, opened 2023-08-08
feat: adding swiftlint

TheRogue76:
Lint consistency in iOS is a pain at the moment, and frankly we should have one uniform approach to stuff anyway. This PR adds the usage of Swiftlint, a popular linting tool for the code base. It adds it as a script to lint it from the JS side.

You need to have Swiftlint installed on your machine (updated the contributing guide) and can do it with:
```bash
brew install swiftlint
```
The ideal next step is probably a small github action that has Swiftlint, running it on PR commits.


package.json
```diff
@@ -11,6 +11,7 @@
     "fabric:ios": "yarn workspace fabric-example ios",
     "paper:android": "yarn workspace paper-example android",
     "paper:ios": "yarn workspace paper-example ios",
+    "lint:ios": "cd packages/core/ios && swiftlint --quiet --fix && swiftlint --quiet && cd ../../..",
```
matinzd, line 14:
```suggestion
    "lint:ios": "swiftlint --quiet --fix ./packages/core/ios && swiftlint --quiet ./packages/core/ios",
```

CONTRIBUTING.md
```diff
@@ -56,6 +56,19 @@ codebase, however you can always check to see if the source code is compliant by
...
+For linting the native iOS package, we are using [Swift lint](https://github.com/realm/SwiftLint). You need to install it on your machine using the following command:
+
+```bash
+brew install swiftlint
+```
+
+And then you can run it by calling it from JS using:
+
+```bash
+yarn lint:ios
```
matinzd, line 68: I am getting a lot of errors and warninigs.

```
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/AnimationViewManagerModule.swift:36:1: warning: Line Length Violation: Line should be 120 characters or less; currently it has 131 characters (line_length)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/PlatformColor.swift:19:9: error: Identifier Name Violation: Variable name 'r' should be between 3 and 40 characters long (identifier_name)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/PlatformColor.swift:19:9: error: Identifier Name Violation: Variable name 'g' should be between 3 and 40 characters long (identifier_name)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/PlatformColor.swift:19:9: error: Identifier Name Violation: Variable name 'b' should be between 3 and 40 characters long (identifier_name)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/PlatformColor.swift:19:9: error: Identifier Name Violation: Variable name 'a' should be between 3 and 40 characters long (identifier_name)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/ContainerView.swift:101:59: error: Force Cast Violation: Force casts should be avoided (force_cast)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/ContainerView.swift:102:58: error: Force Cast Violation: Force casts should be avoided (force_cast)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/ContainerView.swift:309:74: error: Force Cast Violation: Force casts should be avoided (force_cast)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/ContainerView.swift:311:98: error: Force Cast Violation: Force casts should be avoided (force_cast)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/ContainerView.swift:277:1: warning: Line Length Violation: Line should be 120 characters or less; currently it has 130 characters (line_length)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/ContainerView.swift:311:1: warning: Line Length Violation: Line should be 120 characters or less; currently it has 133 characters (line_length)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/ContainerView.swift:359:1: warning: Line Length Violation: Line should be 120 characters or less; currently it has 143 characters (line_length)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/ContainerView.swift:132:13: warning: No Fallthrough only Violation: Fallthroughs can only be used if the `case` contains at least one other statement (no_fallthrough_only)
/Users/matin/work/personal/lottie-react-native/packages/core/ios/LottieReactNative/ContainerView.swift:14:1: warning: Type Body Length Violation: Type body should span 250 lines or less excluding comments and whitespace: currently spans 278 lines (type_body_length)
```

TheRogue76: We can take two approaches here, refactor, or shutting off some of the rules:

1. Number of characters in lines? Maybe appropriate to refactor
2. Variable names length? I think we can disable
3. Force Cast? Generally a big No-No, but i think for exceptional cases, we can add // swiftlint:disable force_cast (We have opt-outs)
4. No fall through? Generally Swift tries to be a lot stricter with Switch cases than TS/JS, so we can refactor
5. Body length violation? Generally, yes, don't write too big of a Swift file, but this might be a bit too strict

What do you think about each case? I can disable things globally with a simple .swiftlint file in the iOS directory, works a lot like .eslint

TheRogue76:
> Maybe it's a good idea to add this to pre-commit hook.

Not a huge fan of pre-commit hooks in general, but might be a good idea in this case. Give this comment a thumbs up and i will add it

packages/core/ios/LottieReactNative.xcodeproj/project.pbxproj
```diff
@@ -118,6 +118,7 @@
 				11FA5C4D1C4A1296003AC2EE /* Sources */,
 				11FA5C4E1C4A1296003AC2EE /* Frameworks */,
 				11FA5C4F1C4A1296003AC2EE /* CopyFiles */,
+				423561892A827E2800D7D5AB /* Run SwiftLint */,
```
matinzd, line 121: I don't think this is a good idea to run this on users' environments. This should only be done on our side.

CONTRIBUTING.md
matinzd: Let's add swift lint file and suppress some of them.

packages/core/ios/LottieReactNative.xcodeproj/project.pbxproj
matinzd: Let's do it in CI.

TheRogue76: Makes sense. Removed under [fbbe7dd](https://github.com/lottie-react-native/lottie-react-native/pull/1085/commits/fbbe7ddb5f08e54b6f89ea935e485eb8f3747a08). Gonna add a CI step after the PR is merged

TheRogue76:
End result:

1-Number of lines limitation is off for now, (frankly, i don't see us needing it enough to go through the hassle of enabling it, and it doesn't bring that much value with the amount of code we have)

2-Variable name limitation is off, doesn't make sense to limit our variable names

3-I addressed the force cast issue. @matinzd Please take a look at [b91fb6e](https://github.com/lottie-react-native/lottie-react-native/pull/1085/commits/b91fb6e5287a358f1d6eb254c16b3f70291c4f86) and let me know what you think. TS should stop them from ever making such a mistake, but frankly i rather be safe than sorry

4-I dealt with fall through by setting render mode `automatic` as the default case, along with any other garbled values that they might push through. The better way to do it is to make the value be an enum, not a string, but maybe that is something we can work on next? (The codegen type definitions are marking them as string for both this and the `resizeMode`, but based on https://github.com/reactwg/react-native-new-architecture/discussions/91#discussioncomment-5904287, we might be able to just give it a basic string union and expect it to work)

5- Much like number 1, i feel like while it helps improve readability, being too stringy with it is kind of dumb. The new rule set gives warnings at 300 lines and above for the size of the body, and at 400, it will turn into an error, asking the creator to refactor their code. I feel like we can test this for now and figure out along the way, but considering we can write extensions, i don't think it is that big of an issue

matinzd:
1. Agree with that
2. We don't need a lot of restrictions here
3. Have you tried testing color filters and text filters prop? Is it still working without an issue?
4. Enum is not possible for now. We can think of it later.

matinzd:
I separated two scripts one for linting and the other for fixing lint issues.

PR #1087, opened 2023-08-08
chore(ci): add ci for ios example build test

matinzd:
Add CI to test against different versions of XCode starting from 13.3.1.

Docs for macos-12 runners: https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

### DIFF
--- a/.github/workflows/ios-build.yml
+++ b/.github/workflows/ios-build.yml
@@ -1,6 +1,4 @@
-name: iOS Build
-
-# iOS matrix build on XCode 13.3 up to latest
+name: Build iOS Paper Example
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


TheRogue76:
Man after my own heart

.github/workflows/ios-build.yml
```diff
@@ -0,0 +1,40 @@
...
+  build-and-test:
+    runs-on: macos-latest
+
+    strategy:
+      matrix:
+        xcode-version: [13.3, latest]
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
```
TheRogue76, line 23: Use V3

TheRogue76:
Does this mean we will also need a macos 13 machine to run the latest stuff? Cause 12 stops at Xcode 14.2

Issue #1090, opened 2023-08-14
6.1.2 DotLottie's not working for Android Release mode


walterholohan:
### Description

When building a release version of a RN app, the `.lottie` files will not load.

### Steps to Reproduce

1. react-native run-android --variant=release
2. install release version of app on device

**Expected behavior:** [What you expected to happen]

The lotties should load

**Actual behavior:** [What actually happened]

The lotties will not load, and you get an error in `onAnimationFailure` of `no protocol`

### React Native Environment
```
System:
    OS: macOS 13.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 84.34 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.1 - ~/.nvm/versions/node/v18.16.1/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.1/bin/npm
    Watchman: 2023.07.10.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.12.1 - /opt/homebrew/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.4, iOS 16.4, macOS 13.3, tvOS 16.4, watchOS 9.4
    Android SDK: Not Found
  IDEs:
    Android Studio: 2022.2 AI-222.4459.24.2221.9862592
    Xcode: 14.3.1/14E300c - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.12 - /usr/local/opt/openjdk@11/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.2.0 => 18.2.0 
    react-native: 0.71.12 => 0.71.12 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

`metro.config.js`
```
/**
 * Metro configuration for React Native
 * https://github.com/facebook/react-native
 *
 * @format
 */
const { getDefaultConfig } = require('metro-config')

module.exports = (async () => {
	const {
		resolver: { sourceExts, assetExts },
	} = await getDefaultConfig()
	return {
		transformer: {
			babelTransformerPath: require.resolve('react-native-svg-transformer'),
			getTransformOptions: async () => ({
				transform: {
					experimentalImportSupport: false,
					inlineRequires: true,
				},
			}),
		},
		resolver: {
[...]

PR #1091, opened 2023-08-15
fix: dot lottie for android in release mode

matinzd:
It seems that bundler flattens folder structure using underscore for each subfolder, removes the file extension and the full path in release mode.

Fixes: https://github.com/lottie-react-native/lottie-react-native/issues/1090

Reference: https://github.com/facebook/react-native/issues/24963


TheRogue76 (approved):
I was wondering about the whole loading thing as well. Very nice

walterholohan (approved):

PR #1107, opened 2023-09-22
feat: add web support

kosmydel:
## Rationale
We want to support the `web` platform, to have one library for all platforms.

## Related issues
https://github.com/Expensify/App/issues/26871

## Testing checklist
- [x] Run a `.lottie` animation
- [x] Run a `.json` Lottie animation
- [x] Test `play()`, `pause()`, `reset()`, `resume()` methods.
- [x] Test `source`, `speed`, `loop`, `autoPlay` props.
- [x] Test `onAnimationLoop`, `onAnimationFinish`, `onAnimationFailure` events.
- [x] Test `hover` and `direction` new props.

## Demo video

https://github.com/lottie-react-native/lottie-react-native/assets/104823336/65ef2de6-e274-4780-9050-0f22b700e38b


kosmydel:
It is ready for the review 😇 

cc @roryabraham, @matinzd

packages/core/src/LottieView/index.web.tsx
```diff
@@ -0,0 +1,118 @@
+import React, { forwardRef, useImperativeHandle, useRef, useCallback } from 'react';
+import type { LottieViewProps } from '../types';
+import { parsePossibleSources } from './utils';
+
+let DotLottiePlayer, Player;
+try {
+    DotLottiePlayer = require('@dotlottie/react-player').DotLottiePlayer;
+    Player = require('@lottiefiles/react-lottie-player').Player;
+} catch (e) { }
```
matinzd, line 9: Isn't it better to add warning here as well to make sure they have installed the required deps?

matinzd: Ah my bad. It's being handled down here in this file.

packages/core/src/LottieView/index.web.tsx
```diff
@@ -0,0 +1,118 @@
...
+      onAnimationFailure,
+      onAnimationFinish,
+      onAnimationLoop,
+    }: LottieViewProps,
+    ref: any,
+  ) => {
+    const sources = parsePossibleSources(source);
+    const isLottie =
+      sources.sourceName?.includes('.lottie') || !!sources.sourceDotLottieURI;
+    const lottieSource = sources.sourceDotLottieURI || sources.sourceName;
```
matinzd, line 31: Would you please try loading json file from remote url? 

Try this URL: https://raw.githubusercontent.com/lottie-react-native/lottie-react-native/master/apps/paper/src/animations/Watermelon.json

Tried disabling CORS but not luck.

kosmydel: Fixed!

matinzd:
An edge case here:

It seems there is something wrong when the source is invalid then switching to local json file. The lib cannot recover the new source. 
If I switch from invalid URL to dot lottie the source can recover an it starts playing.

matinzd:
Are we going to have a second PR for this? e.g #2? If not please remove "(#1)" from the title.
Thanks for your help and your contribution!

kosmydel:
> An edge case here:
> 
> It seems there is something wrong when the source is invalid then switching to local json file. The lib cannot recover the new source. If I switch from invalid URL to dot lottie the source can recover an it starts playing.
> 
>  d.mov

I added a solution to that. But it might be better to solve this problem in the used library.

matinzd (approved):
Good to go! 🚀

tvisentin:
🥳

packages/core/src/LottieView/index.web.tsx
```diff
@@ -0,0 +1,134 @@
...
+    if (isLottie) {
+      if (!DotLottiePlayer) {
+        throw new Error('lottie-react-native: The module @dotlottie/react-player is missing.');
+      }
+      return (
+        <DotLottiePlayer
+          lottieRef={playerRef}
+          src={lottieSource}
+          onEvent={handleEvent}
+          style={webStyle}
```
kosmydel, line 104: We might consider passing here simply `style`. But the library uses the type `React.CSSProperties`, not `ViewStyle` from the `react-native`. I am thinking about if and how to solve it.

cc @matinzd, @blazejkustra

matinzd: Let's keep `React.CSSProperties`.

PR #1120, opened 2023-10-16
feat: Add callback for when animation has loaded

thevoiceless:
This is part feature, part bug report.

I want to use the imperative API to set the initial frame to display after loading a `.lottie` file. The following worked as expected on Android, but not on iOS; as far as I can tell, the imperative call is being ignored because it happens before the `.lottie` has finished loading:
```ts
function App() {
  const animationRef = React.useRef<LottieView>(null);

  React.useEffect(() => {
    animationRef?.current?.play(100, 100);
  }, []);

  return (
    <LottieView
      ref={animationRef}
      source={require('./animation.lottie')}
      style={{width: 300, height: 300}}
    />
  );
}
```
 I personally would consider the differing behavior to be a bug, and lean towards the "it just works" result on Android. 

Instead of assuming the intended behavior, I opted to add an `onAnimationLoaded` callback. I used a barebones React Native app to verify that it works on the old and new architecture.


matinzd:
Please resolve conflicts with the main branch.

packages/core/ios/LottieReactNative/ContainerView.swift
```diff
@@ -295,6 +309,10 @@ class ContainerView: RCTView {
 
         applyColorProperties()
         playIfNeeded()
+
+        animationView?.animationLoaded = { animationView, animation in
```
matinzd, line 313:
```suggestion
        animationView?.animationLoaded = { [weak self] animationView, animation in
```

TheRogue76: To follow up, after the weak self, use something like:
```swift
guard let self = self else { return }
```
To keep in line with the way we handle reference cycles. It also let's you keep the rest of your code as is

TheRogue76 (commented):
Looks good. Just needs to break that one reference cycle and we can merge it

TheRogue76 (approved):

matinzd (approved):

PR #1127, opened 2023-10-30
fix(web): don't require @dotlottie/react-player to be installed for JSON animations

adamhari:
The current web implementation requires projects to include both `@lottiefiles/react-lottie-player` and `@dotlottie/react-player` because they are required in the same try/catch block.

Apps should be able to install either of these depending on whether they need to play `.lottie` files or not.

This PR separates the requires into separate try/catch blocks so `@lottiefiles/react-lottie-player` can be installed and used without including `@dotlottie/react-player`.


matinzd (approved):
Thanks for catching this!

kosmydel:
Thank you for noticing and catching this! 
I've just also spotted this issue and wanted to fix it, but you were first :)

PR #1134, opened 2023-11-27
feat(android): support local file path

jhen0409:
~~iOS:~~
~~- Check uri prefix to use `filePath:` or `dotLottieFilePath:`~~

Android:
- Check uri prefix to use `FileInputStream` or `ZipInputStream`
- lottie-android v6.2 includes https://github.com/airbnb/lottie-android/pull/2411 so we can use ZipInputStream for load local dotlottie file

Fixes: https://github.com/lottie-react-native/lottie-react-native/issues/1142


jhen0409:
Alright, I've found that iOS may not require these additional changes as it works by default. However, we only need to implement these changes for Android to support local file.

packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
```diff
@@ -119,13 +123,31 @@ class LottieAnimationViewPropertyManager(view: LottieAnimationView) {
         }
 
         animationURL?.let {
-            view.setAnimationFromUrl(it, it.hashCode().toString())
+            if (it.startsWith('/') || it.startsWith("file://")) {
+                try {
+                    view.setAnimation(FileInputStream(File(it)), Integer.toString(it.hashCode()))
```
matinzd, line 128: Is `Integer.toString(it.hashCode())` for caching purposes?

packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
```diff
@@ -119,13 +123,31 @@ class LottieAnimationViewPropertyManager(view: LottieAnimationView) {
         }
 
         animationURL?.let {
-            view.setAnimationFromUrl(it, it.hashCode().toString())
+            if (it.startsWith('/') || it.startsWith("file://")) {
```
matinzd, line 126: Can you give me an example for starting with `/` ? I feel like this checking can be unreliable.

jhen0409: I think `startsWith('/')` is should reliable for unix file system. We could use `File.exists()` instead, which should be cleaner.

matinzd: Yes! Much better :) Thanks.

packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
```diff
@@ -24,6 +24,10 @@ import com.facebook.react.views.text.TextAttributeProps.UNSET
 import com.facebook.react.util.RNLog
 import java.lang.ref.WeakReference
 import java.util.regex.Pattern
+import java.util.zip.ZipInputStream
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileNotFoundException
```
matinzd, line 30: We can remove this I guess.

matinzd (approved):

matinzd:
Thanks for your contribution!

Issue #1067, opened 2023-07-27
Lottie and Fabric/New Arch still fails on 6.0.1


billnbell:
I updated the lottie-react-native.podspec with a patch - and this now works for me:

```
require 'json'

package = JSON.parse(File.read(File.join(__dir__, "package.json")))

folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'

Pod::Spec.new do |s|
  s.name                    = package["name"]
  s.version                 = package['version']
  s.summary                 = package["description"]
  s.license                 = package["license"]

  s.author                  = package["author"]
  s.homepage                = package["homepage"]

  s.ios.deployment_target   = '12.4'
  s.osx.deployment_target   = '10.10'
  s.tvos.deployment_target  = '12.4'

  s.source                  = { :git => "https://github.com/lottie-react-native/lottie-react-native.git", :tag => "v#{s.version}" }
  s.source_files            = "ios/**/*.{h,m,mm,swift}"

  s.dependency 'lottie-ios', '~> 4.2.0'

  if ENV['RCT_NEW_ARCH_ENABLED'] == '1' then
    s.compiler_flags = folly_compiler_flags + " -DRCT_NEW_ARCH_ENABLED=1"
    s.pod_target_xcconfig    = {
      "HEADER_SEARCH_PATHS" => "\"$(PODS_ROOT)/boost\"",
      "OTHER_CPLUSPLUSFLAGS" => "-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1",
      "CLANG_CXX_LANGUAGE_STANDARD" => "c++17",
      "DEFINES_MODULE" => "YES",
    }
    
    s.dependency "React-RCTFabric"
    s.dependency "React-Codegen"
    s.dependency "RCT-Folly"
    s.dependency "RCTRequired"
    s.dependency "RCTTypeSafety"
    s.dependency "ReactCommon/turbomodule/core"
    s.dependency "React-utils"
    s.subspec "xxxutils" do |ss|
      ss.dependency "ReactCommon"
      ss.dependency "React-utils"
      ss.source_files         = "react/utils/**/*.{cpp,h}"
      ss.header_dir           = "react/utils"
[...]

Issue #1138, opened 2023-12-12
Build issue in react-native v0.73 new architecture.


neeteshraj:
### Description
The following build commands failed:
	CompileC /Users/niteshrajkhanal/Library/Developer/Xcode/DerivedData/singlecustomerapp-ggsyyaqexacwdkfzzmjannnihwfm/Build/Intermediates.noindex/Pods.build/Debug-iphoneos/lottie-react-native.build/Objects-normal/arm64/LottieAnimationViewComponentView.o /Users/niteshrajkhanal/Desktop/development/sca-mobile-app/node_modules/lottie-react-native/ios/Fabric/LottieAnimationViewComponentView.mm normal arm64 objective-c++ com.apple.compilers.llvm.clang.1_0.compiler (in target 'lottie-react-native' from project 'Pods')
(1 failure)

### Steps to Reproduce

1. Create a new react native application.
2. Package.json 
{
  "name": "singlecustomerapp",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "android": "react-native run-android",
    "ios": "react-native run-ios",
    "ios-15": "npx react-native run-ios --simulator='iPhone 15 Pro'",
    "lint": "eslint .",
    "start": "react-native start",
    "test": "jest",
    "type-check": "tsc",
    "test:report": "jest --collectCoverage --coverageDirectory=\"./coverage\" --ci --reporters=default --reporters=jest-junit --coverage",
    "pod-install": "cd ios && RCT_NEW_ARCH_ENABLED=1 bundle exec pod install && cd ..",
    "gen-release-apk": "cd android && ./gradlew assembleRelease && cd ..",
    "gen-release-bundle": "cd android && ./gradlew bundleRelease && cd ..",
    "bundle-android": "react-native bundle --platform android --dev false --entry-file index.js --bundle-output android/app/src/main/assets/index.android.bundle",
    "release-android": "cd android && ./gradlew clean && cd .. && react-native bundle --platform android --dev false --entry-file index.js --bundle-output android/app/src/main/assets/index.android.bundle && cd android && ./gradlew assembleRelease && cd ..",
    "adb-log": "adb logcat *:E",
    "prepare": "husky install"
  },
  "dependencies": {
    "@react-native-masked-view/masked-view": "^0.3.0",
    "@react-navigation/bottom-tabs": "^6.5.11",
    "@react-navigation/native": "^6.1.9",
    "@react-navigation/stack": "^6.3.20",
[...]

PR #1139, opened 2023-12-19
chore(iOS): Use `install_modules_dependencies` in Podspec

cipolleschi:
This change updates the `lottie-react-native.podspec` to consume the `install_modules_dependencies` function provided by React Native to configure the pods dependencies.
Thanks to using this function, whenever we update the pods structure in React Native, the library will automatically benefit from it.
This will make the library more future proof and more resilient to changes in React Native.

This allow Lottie to work also with `use_frameworks!` and with the New Architecture. Before the change, it failed to build when `use_frameworks!` was set in RN 0.73.x

## Test plan

Tested locally:

Before changes
1. Created a new app with `npx react-native init RNFrameworks --version
latest --skip-install`
1. `cd RNFrameworks`
1. `yarn add lottie-react-native`
1. `yarn install`
1. `cd ios`
1. `NO_FLIPPER=1 USE_FRAMEWORKS=dynamic RCT_NEW_ARCH_ENABLED=1 bundle
exec pod install`
1. `open RNFrameworks`
1. build and run and observe it failing

Apply the changes to local node_modules's `lottie-react-native.podspec` changes
1. Reinstall pods `NO_FLIPPER=1 USE_FRAMEWORKS=dynamic
RCT_NEW_ARCH_ENABLED=1 bundle exec pod install`
1. `open RNFrameworks`
1. build and run and observe it succeeding

Fixes: #1067, #1138

### DIFF
--- a/packages/core/lottie-react-native.podspec
+++ b/packages/core/lottie-react-native.podspec
@@ -2,8 +2,6 @@ require 'json'
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
 
-folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'
-
 Pod::Spec.new do |s|
   s.name                    = package["name"]
   s.version                 = package['version']
@@ -21,26 +19,35 @@ Pod::Spec.new do |s|
   s.source_files            = "ios/**/*.{h,m,mm,swift}"
 
   s.dependency 'lottie-ios', '~> 4.3.3'
-  
+
   s.swift_version = '5.6'
 
-  if ENV['RCT_NEW_ARCH_ENABLED'] == '1' then
-    s.compiler_flags = folly_compiler_flags + " -DRCT_NEW_ARCH_ENABLED=1"
-    s.pod_target_xcconfig    = {
-      "HEADER_SEARCH_PATHS" => "\"$(PODS_ROOT)/boost\"",
-      "OTHER_CPLUSPLUSFLAGS" => "-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1",
-      "CLANG_CXX_LANGUAGE_STANDARD" => "c++17",
-      "DEFINES_MODULE" => "YES",
-    }
-    
-    s.dependency "React-RCTFabric"
-    s.dependency "React-Codegen"
-    s.dependency "RCT-Folly"
-    s.dependency "RCTRequired"
-    s.dependency "RCTTypeSafety"
-    s.dependency "ReactCommon/turbomodule/core"
+  # install_modules_dependencies has been defined since React Native 70
+  # if you only support React Native 70+, we can remove the `if defined?() else` statement
+  if defined?(install_modules_dependencies)
+    install_modules_dependencies(s)
+    if !(ENV['RCT_NEW_ARCH_ENABLED'] == '1') then
+      s.exclude_files = "ios/Fabric"
+    end
   else
-    s.dependency 'React-Core'
-    s.exclude_files = "ios/Fabric"
+    if ENV['RCT_NEW_ARCH_ENABLED'] == '1' then
+      s.compiler_flags = folly_compiler_flags + " -DRCT_NEW_ARCH_ENABLED=1"
+      s.pod_target_xcconfig    = {
+        "HEADER_SEARCH_PATHS" => "\"$(PODS_ROOT)/boost\"",
+        "OTHER_CPLUSPLUSFLAGS" => "-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1",
+        "CLANG_CXX_LANGUAGE_STANDARD" => "c++17",
+        "DEFINES_MODULE" => "YES",
+      }
+
+      s.dependency "React-RCTFabric"
+      s.dependency "React-Codegen"
+      s.dependency "RCT-Folly"
+      s.dependency "RCTRequired"
+      s.dependency "RCTTypeSafety"
+      s.dependency "ReactCommon/turbomodule/core"
+    else
+      s.dependency 'React-Core'
+      s.exclude_files = "ios/Fabric"
+    end
   end
 end
--- a/packages/core/lottie-react-native.podspec
+++ b/packages/core/lottie-react-native.podspec
@@ -31,6 +31,7 @@ Pod::Spec.new do |s|
     end
   else
     if ENV['RCT_NEW_ARCH_ENABLED'] == '1' then
+      folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'
       s.compiler_flags = folly_compiler_flags + " -DRCT_NEW_ARCH_ENABLED=1"
       s.pod_target_xcconfig    = {
         "HEADER_SEARCH_PATHS" => "\"$(PODS_ROOT)/boost\"",
--- a/packages/core/ios/Fabric/LottieAnimationViewComponentView.mm
+++ b/packages/core/ios/Fabric/LottieAnimationViewComponentView.mm
@@ -19,18 +19,22 @@ using namespace facebook::react;
     LottieContainerView *_view;
 }
 
++(void) load {
+    [super load];
+}
+
 - (instancetype) initWithFrame:(CGRect)frame
 {
     if (self = [super initWithFrame:frame]) {
         static const auto defaultProps = std::make_shared<const LottieAnimationViewProps>();
         _props = defaultProps;
-        
+
         _view = [LottieContainerView new];
         _view.delegate = self;
-        
+
         self.contentView = _view;
     }
-    
+
     return self;
 }
 
@@ -44,55 +48,55 @@ using namespace facebook::react;
 - (void)updateProps:(Props::Shared const &)props oldProps:(Props::Shared const &)oldProps {
     const auto &oldLottieProps = *std::static_pointer_cast<const LottieAnimationViewProps>(_props);
     const auto &newLottieProps = *std::static_pointer_cast<const LottieAnimationViewProps>(props);
-    
+
     if(oldLottieProps.resizeMode != newLottieProps.resizeMode) {
         [_view setResizeMode:RCTNSStringFromString(newLottieProps.resizeMode)];
     }
-    
+
     if(oldLottieProps.sourceJson != newLottieProps.sourceJson) {
         [_view setSourceJson:RCTNSStringFromString(newLottieProps.sourceJson.c_str())];
     }
-    
+
     if(oldLottieProps.sourceDotLottieURI != newLottieProps.sourceDotLottieURI) {
         [_view setSourceDotLottieURI:RCTNSStringFromString(newLottieProps.sourceDotLottieURI)];
     }
-    
+
     if(oldLottieProps.sourceName != newLottieProps.sourceName) {
         [_view setSourceName:RCTNSStringFromString(newLottieProps.sourceName)];
     }
-    
+
     if(oldLottieProps.sourceURL != newLottieProps.sourceURL) {
         [_view setSourceURL:RCTNSStringFromString(newLottieProps.sourceURL)];
     }
-    
+
     if(oldLottieProps.progress != newLottieProps.progress) {
         [_view setProgress:newLottieProps.progress];
     }
-    
+
     if(oldLottieProps.loop != newLottieProps.loop) {
         [_view setLoop:newLottieProps.loop];
     }
-    
+
     if(oldLottieProps.speed != newLottieProps.speed) {
         [_view setSpeed:newLottieProps.speed];
     }
-    
+
     if(oldLottieProps.colorFilters != newLottieProps.colorFilters) {
         [_view setColorFilters:convertColorFilters(newLottieProps.colorFilters)];
     }
-    
+
     if(oldLottieProps.textFiltersIOS != newLottieProps.textFiltersIOS) {
         [_view setTextFiltersIOS:convertTextFilters(newLottieProps.textFiltersIOS)];
     }
-    
+
     if(oldLottieProps.renderMode != newLottieProps.renderMode) {
         [_view setRenderMode:RCTNSStringFromString(newLottieProps.renderMode)];
     }
-    
+
     if(oldLottieProps.autoPlay != newLottieProps.autoPlay) {
         [_view setAutoPlay:newLottieProps.autoPlay];
     }
-    
+
     [super updateProps:props oldProps:oldProps];
 }
 
@@ -132,11 +136,11 @@ using namespace facebook::react;
     if(!_eventEmitter) {
         return;
     }
-    
+
     LottieAnimationViewEventEmitter::OnAnimationFinish event = {
         .isCancelled = isCancelled
     };
-    
+
     std::dynamic_pointer_cast<const LottieAnimationViewEventEmitter>(_eventEmitter)->onAnimationFinish(event);
 }
 
@@ -145,11 +149,11 @@ using namespace facebook::react;
     if(!_eventEmitter) {
         return;
     }
-    
+
     LottieAnimationViewEventEmitter::OnAnimationFailure event = {
         .error = std::string([error UTF8String])
     };
-    
+
     std::dynamic_pointer_cast<const LottieAnimationViewEventEmitter>(_eventEmitter)->onAnimationFailure(event);
 }
 


TheRogue76 (approved):
Looks good. Didn't realise we had an issue with `use_framework` and new arch, @matinzd Maybe we can add something to the CI pipeline to test this out as well?

heath-clink:
I applied this change to my project, and installed pods as shown (`NO_FLIPPER=1 USE_FRAMEWORKS=dynamic RCT_NEW_ARCH_ENABLED=1 bundle exec pod install`)

The project builds and runs without error, but now a LottieView appears as a pink box with "Unimplemented component" in it

TheRogue76:
> I applied this change to my project, and installed pods as shown (`NO_FLIPPER=1 USE_FRAMEWORKS=dynamic RCT_NEW_ARCH_ENABLED=1 bundle exec pod install`)
> 
> The project builds and runs without error, but now a LottieView appears as a pink box with "Unimplemented component" in it.
> 
> UPDATE: If I change `USE_FRAMEWORKS` to `static` it works.

What is your project running? Can you give a `npx react-native info` log? (Mostly curious about RN version)

packages/core/lottie-react-native.podspec
```diff
@@ -2,8 +2,6 @@ require 'json'
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
 
-folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'
```
matinzd, line 5: Any reason on why this is removed? It's being reference later on line34 in the new change.

packages/core/lottie-react-native.podspec
```diff
@@ -21,26 +19,35 @@ Pod::Spec.new do |s|
...
-    s.dependency "RCT-Folly"
-    s.dependency "RCTRequired"
-    s.dependency "RCTTypeSafety"
-    s.dependency "ReactCommon/turbomodule/core"
+  # install_modules_dependencies has been defined since React Native 70
+  # if you only support React Native 70+, we can remove the `if defined?() else` statement
+  if defined?(install_modules_dependencies)
+    install_modules_dependencies(s)
+    if !(ENV['RCT_NEW_ARCH_ENABLED'] == '1') then
+      s.exclude_files = "ios/Fabric"
```
matinzd, line 30: I guess you forgot to to add 'React-Core' dependency here for non fabric (old arch).

```suggestion
      s.dependency 'React-Core'
      s.exclude_files = "ios/Fabric"
```

packages/core/lottie-react-native.podspec
```diff
@@ -21,26 +19,35 @@ Pod::Spec.new do |s|
...
-    s.dependency "React-RCTFabric"
-    s.dependency "React-Codegen"
-    s.dependency "RCT-Folly"
-    s.dependency "RCTRequired"
-    s.dependency "RCTTypeSafety"
-    s.dependency "ReactCommon/turbomodule/core"
+  # install_modules_dependencies has been defined since React Native 70
+  # if you only support React Native 70+, we can remove the `if defined?() else` statement
+  if defined?(install_modules_dependencies)
+    install_modules_dependencies(s)
```
matinzd, line 28: I think `React-Core` should not be included for fabric? 

https://github.com/facebook/react-native/blob/90fb73e83caf087659b6b3173edd457b4f1d1987/packages/react-native/scripts/cocoapods/new_architecture.rb#L108

Sorry I am a bit confused here ...

packages/core/lottie-react-native.podspec
matinzd: Maybe this? https://github.com/facebook/react-native/blob/90fb73e83caf087659b6b3173edd457b4f1d1987/packages/react-native/scripts/cocoapods/new_architecture.rb#L11C7-L11C19

cipolleschi: this can be removed, in theory, because it is already declared in the function `install_modules_dependencies` so you don't need to specify it... but I overlooked line 34.. so we need to restore it. sorry. :(

packages/core/lottie-react-native.podspec
cipolleschi: No, we still need it for the time being. As soon as we will remove the Old Arch, we might remove it.

cipolleschi:
> UPDATE: If I change USE_FRAMEWORKS to static it works.

Ah... that's weird. I'll double-check that. There might be something missing. It is the first time that I see static working and dynamic not working. 🤔

packages/core/lottie-react-native.podspec
matinzd: No worries! I guess that might describe why `dynamic` is not working?

packages/core/lottie-react-native.podspec
```diff
@@ -21,26 +19,35 @@ Pod::Spec.new do |s|
...
+    if !(ENV['RCT_NEW_ARCH_ENABLED'] == '1') then
+      s.exclude_files = "ios/Fabric"
+    end
   else
-    s.dependency 'React-Core'
-    s.exclude_files = "ios/Fabric"
+    if ENV['RCT_NEW_ARCH_ENABLED'] == '1' then
+      s.compiler_flags = folly_compiler_flags + " -DRCT_NEW_ARCH_ENABLED=1"
```
matinzd, line 34:
```suggestion
      folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'

      s.compiler_flags = folly_compiler_flags + " -DRCT_NEW_ARCH_ENABLED=1"
```

heath-clink:
> What is your project running? Can you give a `npx react-native info` log? (Mostly curious about RN version)

RN 0.73.1

Here's the repro project: https://github.com/heath-clink/rn73-lottie-repro

TheRogue76:
> > What is your project running? Can you give a `npx react-native info` log? (Mostly curious about RN version)
> 
> RN 0.73.1
> 
> Here's the repro project: https://github.com/heath-clink/rn73-lottie-repro
> 
> It's just the template project from `react-native init` with lottie added

I'll check today to see what i can find and what might be wrong

TheRogue76:
So, I messed around a bit, this is what i am seeing when using both dynamic and static frameworks on the test repo:

When it is static the first callback is called and the Lottie registers fine:

Screenshot 2023-12-20 at 16 40 11

When i set it to dynamic, `registerComponentIfPossible` is called like before and Lottie's name is shown, but then it fails both to initialize in both fallbacks:

Screenshot 2023-12-20 at 16 35 59

Screenshot 2023-12-20 at 16 36 48

Looking at `RCTThirdPartyFabricComponentsProvider` which is the implementation detail for `RCTComponentViewClassWithName`, Lottie is there, but RCT_DYNAMIC_FRAMEWORKS is false for some unknown reason, so when it goes to search to see if it can find it, it just returns nil

Screenshot 2023-12-20 at 16 51 27

So the question is why is RCT_DYNAMIC_FRAMEWORKS being reported as false? @cipolleschi any ideas?

TheRogue76 (changes requested):
The class is generated for Dynamic, but the flag is being reported as false so this breaks when framework usage is set to dynamic

TheRogue76:
> 1. Thanks to inheritance, your components should already extends the `RCTViewComponentView` class. That class exposes this [`load` method](https://github.com/facebook/react-native/blob/main/packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm#L39-L44). You can just add this to your components:
> 
> ```objc
> + (void) load 
> {
>   [super load];
> ```
> 
> and it should work.
> 
> _or_
> 
> 2. The app can manually register the components using the [`RCTComponentViewFactoryComponentProvider` protocol](https://github.com/facebook/react-native/blob/main/packages/react-native/React/Fabric/Mounting/RCTComponentViewFactory.h#L23-L31).
>    The RCTAppDelegate already configures the app with that ([Bridgeless](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/AppDelegate/RCTAppDelegate.mm#L99) or [Bridge](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/AppDelegate/RCTAppDelegate.mm#L116). The App's AppDelegate only have to override [this method](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/AppDelegate/RCTAppDelegate.mm#L99).
> 
> In your case, the easiest way would be 1.
> 
> Sorry for the long message. Dynamic Frameworks are new to 0.73, honestly, and I didn't had the time to write all the documentation yet. 😞 Most users that needs frameworks are only using the static version, which is ok for now, but we are doing all the preparatory work to support dynamic frameworks, which might be also required by Swift (or at least, it is what Apple recommend to use).

Thank you for that explanation. Yes the install dependencies function is quite nice for handling things and adding the load function under `LottieAnimationViewComponentView` (our implementation of `RCTViewComponentView`) does the trick  🚀:

Screenshot 2023-12-20 at 18 12 38

If you wish, you can add it in your PR or if you are busy i can open up a new PR to add that and we merge all of this no problem. @matinzd heads up

cipolleschi:
Tomorrow I'll be working normally (although with GMT+1 Timezone! 😅 ) so I can take care of it! 😄  
But thanks @TheRogue76  for offering to help! 😉

packages/core/lottie-react-native.podspec
matinzd: Good! When you find some time, let's revert this for backward compatibility 😉

matinzd:
Thank you, @TheRogue76 and @cipolleschi, for all your help!

I will wait for you to update the PR. I need to merge another PR as well and then we can include this in the next release. 

Great initiative! I hope we get a better documentation for library maintainers who doesn't have time to deep dive in on how internals work in RN. Tbh, I am ok with exploring RN's repo and I have done it most of the times but I don't have time anymore to dedicate to go through the internals and figure things out.

TheRogue76:
We understand. That's how bleeding edge usually is. Please let us know when your testing is done and we can get this merged.
On a side note, do you guys have any way of testing whether the native component has shown up correctly? All these variants (New arch, Old arch, Dynamic framework, static framework, etc) mean every change needs to be manually tested to make sure it shows up right, and i am thinking if there is some sort of solution that the core team has come up with to make sure the rendering is done correctly, we can tap into it and expand our CI to just handle all of these on its own 🤔

cipolleschi:
Finished with testing. I tested:
- Old Arch / Static Libraries
- New Arch / Static Libraries
- New Arch / Dynamic Frameworks
It works consistently, so it should be good to be merged.

Thank you all for the support!

matinzd (approved):

TheRogue76 (approved):

kirillzyusko:
Hello @cipolleschi 👋 

I've discovered that if you are using new arch and try to install pods without `USE_FRAMEWORKS` variable defined on RN 0.72, then you'll get:

```bash
[!] The following Swift pods cannot yet be integrated as static libraries:

The Swift pod `lottie-react-native` depends upon `React-Fabric`, `React-graphics`, `React-utils`, and `React-debug`, which do not define modules. To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries), you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies.
```

On RN 0.73 everything works fine. So I have few questions:
- would you mind to explain why the error happens on RN 0.72 (and I guess older versions as well?) and not happening on RN 0.73?
- can this change (usage of `install_modules_dependencies`) be treated as a breaking change? (because default `pod install` command will fail and you have to specify `USE_FRAMEWORKS` variable on RN < 0.73).

Thank you in advance 🙌

cipolleschi:
Hi! You don't have to use use_frameworks! at all costs. That is an extra setup that we are offering.

React Native 0.72 only support static frameworks. Vanilla React Native does not build with dynamic frameworks in 0.72. Support for dynamic frameworks have been added in 0.73.

The change is not breaking: your customer that were using static libraries and static frameworks will continue being able to do that. From 0.73 onwards, they can also use dynamic frameworks.

But again, if they don't need to use libraries that require frameworks (é.g.: Firebase force the users to use static frameworks) they can keep using static libraries and the change is transparent.

kirillzyusko:
> The change is not breaking: your customer that were using static libraries and static frameworks will continue being able to do that. From 0.73 onwards, they can also use dynamic frameworks.

@cipolleschi okay, but why plain `RCT_NEW_ARCH_ENABLED=1 pod install` produces this error on RN 0.72?

```sh
[!] The following Swift pods cannot yet be integrated as static libraries:

The Swift pod `lottie-react-native` depends upon `React-Fabric`, `React-graphics`, `React-utils`, and `React-debug`, which do not define modules. To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries), you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies.
```

If I'm using `USE_FRAMEWORKS=static RCT_NEW_ARCH_ENABLED=1 pod install` then pods can be installed successfully on RN 0.72. Do you any ideas why it's happening? 👀 

To check that you can simply revert changes to your commit and run `yarn ios` from `apps/fabric`.

TheRogue76:
@kirillzyusko I just instantiated a 0.72 project with the latest version of lottie (6.4.1) ([link here](https://github.com/TheRogue76/test72lottie)) which does not contain this change. Running a normal `RCT_NEW_ARCH_ENABLED=1 pod install` works just fine.
If what you have done is clone the repo and just revert @cipolleschi's specific commit, chances are you still pulled https://github.com/lottie-react-native/lottie-react-native/pull/1143 which updates fabric example to 73 (I am beginning work on bridgeless and needed to update the fabric example to get things going)

kirillzyusko:
> I just instantiated a 0.72 project with the latest version of lottie (6.4.1) ([link here](https://github.com/TheRogue76/test72lottie)) which does not contain this change. Running a normal RCT_NEW_ARCH_ENABLED=1 pod install works just fine.

@TheRogue76 But eventually you are going to publish a new version soon (let's say `6.5.0`) - for people who is using 0.72 + Fabric + new library version (`6.5.0`) will it be a breaking change to add `USE_FRAMEWORKS` flag in order to install pods? Because otherwise installation seems to fail (at least in your example app at https://github.com/lottie-react-native/lottie-react-native/commit/fb7df2d6cabdbeabb6b34c80ff5a672719d3ae32 commit).

> If what you have done is clone the repo and just revert @cipolleschi's specific commit, chances are you still pulled https://github.com/lottie-react-native/lottie-react-native/pull/1143 which updates fabric example to 73 (I am beginning work on bridgeless and needed to update the fabric example to get things going)

I cloned `lottie-react-native` repo, did a hard reset to https://github.com/lottie-react-native/lottie-react-native/commit/fb7df2d6cabdbeabb6b34c80ff5a672719d3ae32 commit and then tried to run the fabric example app and got the error from my previous message.

cipolleschi:
Hi guys! Sorry for the troubles, but it will make your job maintaining the library much easier in the long run.

I made a quick test too and I can confirm that, yes, on 0.72, the regular `bundle exec pod install` with the new architecture doesn't work.

As @TheRogue76 correctly deduced, it is because your library uses Swift. And Swift requires that all the dependencies defines modules. Something that in 0.72 wasn't true for all the pods exposed by React Native.
We fixed it in 0.73. 

To make the change not breaking, I'm going to open a PR on 0.72 with the fixes (they are just 4 lines). So that lottie will work seamlessly with 0.72 and 0.73.

Out of curiosity, @kirillzyusko @TheRogue76  which versions you would like to support? 
React Native support window is 0.71, 0.72 and 0.73. Do you want to support more versions

kirillzyusko:
@cipolleschi definitely I'd like to see a support for 0.71-0.73 🙂 

I think there is no sense to add it to RN 0.70 since `install_modules_dependencies` was added only in 0.71 (correct me if I'm wrong, I see https://github.com/facebook/react-native/commit/82e9c6ad611f1fb816de056ff031716f8cb24b4e was added in 0.71 tag for the first time):

image

Also @TheRogue76 @cipolleschi I will be glad to hear your thoughts on adding support for older versions - whether you see any sense in it or not 👀

cipolleschi:
Create a PR: https://github.com/facebook/react-native/pull/42075

> Also @TheRogue76 @cipolleschi I will be glad to hear your thoughts on adding support for older versions - whether you see any sense in it or not 👀

Personally, I don't think that there is the need to support older versions of React Native. We ask to people to migrate away from unsupported versions quite aggressively.

> I think there is no sense to add it to RN 0.70 since install_modules_dependencies was added only in 0.71 (correct me if I'm wrong, I see https://github.com/facebook/react-native/commit/82e9c6ad611f1fb816de056ff031716f8cb24b4e was added in 0.71 tag for the first time):

Following from above, 0.70 is out of support, so I would not add support for it. And yes, it was not present in 0.70.
So If you want to support > 0.71 we might also simplify the podspec even further and removing the `if defined?` statement.

TheRogue76:
Hi @cipolleschi and @kirillzyusko

For my part, i believe having the same support window as RN would be nice but not always possible (@matinzd please feel free to pitch in your thoughts as well). In this specific case, considering the script was added from 70, we decided to limit it's usage to 73 and above, since 70 falls out of the support window and we didn't A) know how much effort it would be to fix it even for the versions in the support window and B) Didn't want to break it for 70 users (we get odd requests here and there from people migrating slowly, we tell them to update, but the goal is to not have a breaking change unless we bump major version and tell people it is breaking).

I think adding support for this in the supported window of RN versions will be great if possible. We can also adjust our script to opt in for version 71 and above or simplify and fully opt-in, ignoring 0.70 users, but that's just internals of Lottie at that point and we will probably need to have separate talks about how to do it in a nice way.

Thank you to all of you for the support 🚀

matinzd:
Hey all! 

Thanks for all the discussions and contributions around this topic.

@cipolleschi  I would really love to see a proper fix for that, feel free to open a pull request for that. However for now, we just disabled running this module for RN 72 and below to ensure backward compatibility. You can see this PR for more info:
https://github.com/lottie-react-native/lottie-react-native/pull/1146

Regarding the support window, I can say it's not possible for now to limit it to the same version as RN because there are a lot of legacy apps that needs support and it may create a lot of headaches and noise in our issues.

cipolleschi:
That's fine, and many thanks for being so thoughtful of our user base! 🙇 

As soon as https://github.com/facebook/react-native/pull/42075 is merged and React Native 0.72.8 is released, you should be able to remove those ugly lines. Hopefully, this will happen the first or second week of Jan.

kirillzyusko:
@cipolleschi is there any plans to backport these changes to RN 0.71? 👀

matinzd:
I was going to ask the same thing @kirillzyusko

cipolleschi:
0.71 does not support frameworks! 😅  
https://github.com/reactwg/react-native-new-architecture/discussions/129

I think that the patch will work without issues in 0.71, because I hadn't start any work on pods and use frameworks, so the `install_modules_dependencies` was the exact copy of the old documentation, which should be what you had in the manual pod sequence.

Also, in 0.71 there is no `React-debug` and `React-utils` pod, we split them out of `React-Fabric` in 0.72, IIRC.

I'm making a test, out of curiosity, but you should be good to go.

kirillzyusko:
Okay @cipolleschi 
Thank you for support and time dedicated to this issue!

PR #1140, opened 2023-12-19
fix: animation frames being incorrect on new arch for the first frame

TheRogue76:
The frame of the child animation view on iOS is not setup or updated correctly when we first render the element. This has been due to us reading the `bounds` from `RCTView`, which for some reason is reporting incorrect values on the new arch on initial render. This PR switches us over to overriding and calling `layoutSubiews` directly, which lets us correctly set the frame.

I have tested this on both Fabric and on Paper to make sure nothing has broken for any of the implementations


matinzd:
Finally! 🚀

matinzd (approved):

PR #1143, opened 2023-12-21
chore: upgrade example fabric to 73

TheRogue76:
I want to begin work on bridgeless, but in order to get started, i will need to upgrade the example project to 73 (even if it is just fabric)

Screenshot 2023-12-21 at 20 10 18
Screenshot 2023-12-21 at 20 10 35


matinzd (approved):
Great!

TheRogue76:
@matinzd i don't have merge access, so when you can, please merge as well

PR #1145, opened 2023-12-26
feat: add CI for new arch

TheRogue76:
There was another break in new arch on 0.72 under https://github.com/lottie-react-native/lottie-react-native/pull/1139. We should not be keeping track of all of these manually. This PR ensures that at the very least, any PR merged will need to build for both Fabric and Paper against whatever the current version of RN is (That PR was trying to fix Lottie for the latest version of RN, which we had not migrated to at the time). We should probably expand afterwards and add E2E tests to make sure things show up correctly as well, but one step at a time.


TheRogue76:
I am putting this in draft mode until a fix for version 0.72 and below is found. Limiting this CI to the latest version of Xcode is not a good idea.

TheRogue76:
Marking this PR as ready. @matinzd Please give me your thoughts on anything else we should add. I thought about adding Static and Dynamic framework checks as well, but i didn't go for it

matinzd (approved):

.github/workflows/android-fabric-build.yml
```diff
@@ -0,0 +1,54 @@
+name: Build Android Fabric Example
+
+on:
+  pull_request:
+    branches:
+      - master
```
matinzd, line 6: Let's change all triggers including iOS and Android to manual dispatch. We can run each of them according to the specific PR that needs it. This can also avoid CI costs.

```suggestion
on: workflow_dispatch
```

.github/workflows/ios-fabric-build.yml
```diff
@@ -0,0 +1,57 @@
+name: Build iOS Fabric Example
+
+on:
+  pull_request:
+    branches:
+      - master
```
matinzd, line 6:
```suggestion
on: workflow_dispatch
```

TheRogue76:
@matinzd Ready for your review and merge. I can't run it until it is merged apparantly

matinzd (approved):

matinzd:
Let's go 🚀

minhchienwikipedia:
> ```
> [!] [Codegen] warn: using experimental new codegen integration
> [!] The following Swift pods cannot yet be integrated as static libraries:
> 
> The Swift pod `lottie-react-native` depends upon `glog`, which does not define modules. To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries), you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies.
> Error: Process completed with exit code 1.
> ```
> 
> Would you look at that. It breaks on Xcode 14 as well. This is going to be one happy release.

@TheRogue76  I got this issue, how can I fix it quickly?

TheRogue76:
> > ```
> 
> > [!] [Codegen] warn: using experimental new codegen integration
> 
> > [!] The following Swift pods cannot yet be integrated as static libraries:
> 
> > 
> 
> > The Swift pod `lottie-react-native` depends upon `glog`, which does not define modules. To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries), you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies.
> 
> > Error: Process completed with exit code 1.
> 
> > ```
> 
> > 
> 
> > Would you look at that. It breaks on Xcode 14 as well. This is going to be one happy release.
> 
> 
> 
> @TheRogue76  I got this issue, how can I fix it quickly?

Hi there @minhchienwikipedia .
Update to Xcode 14.3.1 at minimum. I'd recommend going all the way to the current most recent version of Xcode, but 14.3.1 ended up being the oldest working version of Xcode for this particular issue.
What is your package version? Are you using new arch?

ozturkilgim:
> ```
> [!] [Codegen] warn: using experimental new codegen integration
> [!] The following Swift pods cannot yet be integrated as static libraries:
> 
> The Swift pod `lottie-react-native` depends upon `glog`, which does not define modules. To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries), you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies.
> Error: Process completed with exit code 1.
> ```
> 
> Would you look at that. It breaks on Xcode 14 as well. This is going to be one happy release.

Same issue

alainib:
me too. any fix please ?

PR #1146, opened 2023-12-26
fix(workaround): enable `install_modules_dependencies` just for RN 73 and above

matinzd:
`install_modules_dependencies` is not working as expected with `RCT_NEW_ARCH_ENABLED=1` for some reason. This is a temporary fix for 72 and below to ensure backward compatibility.

### DIFF
--- a/packages/core/lottie-react-native.podspec
+++ b/packages/core/lottie-react-native.podspec
@@ -2,6 +2,11 @@ require 'json'
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
 
+# TODO(workaround): We need to replace this line with   if defined?(install_modules_dependencies)
+react_package_file = File.expand_path("../react-native/package.json", File.dirname(__FILE__))
+react_package = JSON.parse(File.read(react_package_file))
+minor_react_version = react_package['version'].split('.')[1].to_i
+
 Pod::Spec.new do |s|
   s.name                    = package["name"]
   s.version                 = package['version']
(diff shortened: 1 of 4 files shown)


packages/core/lottie-react-native.podspec
```diff
@@ -24,7 +29,9 @@ Pod::Spec.new do |s|
 
   # install_modules_dependencies has been defined since React Native 70
   # if you only support React Native 70+, we can remove the `if defined?() else` statement
-  if defined?(install_modules_dependencies)
+  # TODO: checking RN version for backward comaptibility reasons since lottie is not compatible with RN 72 and below with install_modules_dependencies
```
matinzd, line 32: If you have better suggestions for comments here let me know

@TheRogue76

packages/core/lottie-react-native.podspec
```diff
@@ -2,6 +2,11 @@ require 'json'
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
 
+# Read the React Native version from the package.json relative to the podspec file
+react_package_file = File.expand_path("../react-native/package.json", File.dirname(__FILE__))
```
TheRogue76, line 6: I wonder if this would work for monorepos

packages/core/lottie-react-native.podspec
```diff
@@ -24,7 +29,9 @@ Pod::Spec.new do |s|
 
   # install_modules_dependencies has been defined since React Native 70
   # if you only support React Native 70+, we can remove the `if defined?() else` statement
-  if defined?(install_modules_dependencies)
+  # TODO(workaround): checking RN version for backward comaptibility reasons since lottie is not compatible with RN 72 and below with install_modules_dependencies
+  # TODO: We need to replace this line with   if defined?(install_modules_dependencies)
+  if minor_react_version >= 73
```
TheRogue76, line 34: Maybe check both?
```suggestion
  if minor_react_version >= 73 && defined?(install_modules_dependencies)
```

TheRogue76: I changed my mind. Do not do the &&. If this method is ever removed, the fallback won't work either, it will silently fail, better to just let it fail loud and have someone open an issue

TheRogue76 (approved):
Looks good. Let me patch 0.72 with this

packages/core/lottie-react-native.podspec
matinzd: I know. Trying to wrap my head around it. It may not work if the directory is symlinked.

TheRogue76:
Screenshot 2023-12-26 at 17 39 00

72 works fine

packages/core/lottie-react-native.podspec
matinzd: Let's default it to true for now. If someone complains for monorepo, we will come back.

PR #1128, opened 2023-10-31
feat(android): support font assets

jhen0409:
Lottie Android v6 includes feature https://github.com/airbnb/lottie-android/pull/2103 so we can use ReactFontManager to get loaded font assets correctly.

For example, https://lottiefiles.com/animations/3d-loop-button-with-text-change-option-mbnBq8lzk2 will caused app crash during error `Font asset not found fonts/Roboto.ttf`.

This PR:
- If found the font, it should displayed text correctly
- It can avoid the app crash with default typeface fallback if font asset not found

I've been using this patch internally from lottie-react-native v5 (it was Java version) some time. It also fix #961.

For iOS, it should be supported by default.

PR #1151, opened 2024-01-04
fix: remove incorrect typeface default

TheRogue76:
Introduced in https://github.com/lottie-react-native/lottie-react-native/pull/1128, this PR aims to remove the incorrect default that was being set for the type face. The value `400` is not a valid type face, and strictly not required in this context as `getTypeface` always returns a valid type face.

Note for the future: This variant of `ReactFontManager` is now depracated. RN 73 introduces a new package: `com.facebook.react.common.assets.ReactFontManager`. We should consider doing a breaking change at some point to cover both this and the updated podspec build script.

Screenshot 2024-01-04 at 15 46 08


matinzd (approved):

matinzd:
Let's do it. I feel like we are missing a lot with the current setup. I also need to add android fabric example in CI pipeline as well.

Issue #1150, opened 2024-01-04
Build error - WrongConstant - Gradle - React Native build for android


leonardo-coutinho-dev:
I have the following error when trying to build a react native (bare workflow) app for android (cd android && ./gradlew build):

``` node_modules\lottie-react-native\android\src\main\java\com\airbnb\android\react\lottie\LottieAnimationViewPropertyManager.kt ```

**lines 72 & 87 - 2 errors**

```
C:\Projetos\hph-services-app\node_modules\lottie-react-native\android\src\main\java\com\airbnb\android\react\lottie\LottieAnimationViewPropertyManager.kt:72: Error: Must be one of: Typeface.NORMAL, Typeface.BOLD, Typeface.ITALIC, Typeface.BOLD_ITALIC [WrongConstant]
                return typeface?: Typeface.defaultFromStyle(400)
                                                            ~~~
C:\Projetos\hph-services-app\node_modules\lottie-react-native\android\src\main\java\com\airbnb\android\react\lottie\LottieAnimationViewPropertyManager.kt:87: Error: Must be one of: Typeface.NORMAL, Typeface.BOLD, Typeface.ITALIC, Typeface.BOLD_ITALIC [WrongConstant]
                return typeface?: Typeface.defaultFromStyle(400)
                                                            ~~~

   Explanation for issues of type "WrongConstant":
   Ensures that when parameter in a method only allows a specific set of
   constants, calls obey those rules.
```
## Problem:

The **lint error** points to the following line in the **LottieAnimationViewPropertyManager.kt** file:

```
return typeface?: Typeface.defaultFromStyle(400)
```

### The error message states:

```
Error: Must be one of: Typeface.NORMAL, Typeface.BOLD, Typeface.ITALIC, Typeface.BOLD_ITALIC [WrongConstant]
```

I have:

Gradle version: 7.6.2

```
"react-native": "0.72.6",
"lottie-react-native": "^6.4.1"
```

eTryp:
I managed to resolve the compilation issue with a quick fix. It might not be the most elegant solution, but it is effective.
To implement this fix:

1. Open the file at the following path:

> /node_modules/lottie-react-native/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt

2. In the init method of this file, locate and remove the two instances of `Typeface.defaultFromStyle(400).`

TheRogue76:
Hi guys,

[...]

PR #1152, opened 2024-01-05
feat: begin work on adding Ktlint / Android build step to the repo

TheRogue76:
This PR aims to add Ktlint to the repo and maybe setup some extra build steps for Android like we have in iOS. This is an effort to make sure stuff like https://github.com/lottie-react-native/lottie-react-native/issues/1150 can never happen again.


TheRogue76:
Let's see if https://github.com/lottie-react-native/lottie-react-native/actions/runs/7427881570/job/20214305445 fails. If it does, i'll add it to the base build steps in CI and remove the trigger so it runs on every PR

TheRogue76:
It doesn't. gradle assemble straight up ignores it, it's only gradle build that calls lottie-react-native:lint. The linter is ours and already existed as one of our verification gradle tasks, it wasn't a third party linter. @matinzd

TheRogue76:
Scrapping the PR, will do another one, with the linter gradle command only

PR #1158, opened 2024-01-13
fix: setup for rn-windows

Victor0814gui:
## Description
I added only a few basic configurations so that the autolink in react-native-windows recognizes the library. I followed the recommendations from https://microsoft.github.io/react-native-windows/docs/native-modules-setup.
 
## Steps To Reproduce

Follow these steps at https://microsoft.github.io/react-native-windows/docs/getting-started and create a React Native app. Then, run ```npx react-native autolink-windows```. You'll notice that the file LottieReactNative.vsxproj is not present in the solutions list.

## Expected Results

The LottieReactNative.vcxproj file should appear in the list of solutions.


TheRogue76:
Hmmm. I can't run anything locally right now, since I don't have a windows machine and i am a bit hesitant about touching anything in core. I'll run the CI to see if anything breaks with this in our current setup in the meantime and maybe we can update the examples to include windows. @matinzd what do you think?

Victor0814gui:
Good question, I also hesitated to make any changes to the core in an inappropriate way. To keep this as transparent and clear as possible, I will add CI to see if anything breaks on Windows. I believe this is a critical bug, as without it, lottie-react-native 6.x.x doesn't work on Windows.

TheRogue76:
Sounds good. Please add a Windows target to the existing Apps, rather than creating a new app. Keeping both Apps in check has been difficult as is, no need to add a new one

.github/workflows/ci-windows.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: yarn build
+      run: |
+        yarn build
+        yarn tsc
+
+    - name: Build x64 release
+      shell: powershell
+      run: yarn windows --release --no-packager --no-deploy --logging
+
+    # Workaround for a bug in package searching during deploy.
```
TheRogue76, line 37: If it is a known issue, please add a link to it here so that in the future, if anyone wants to refactor, they can check the link for status updates

TheRogue76:
The CI failed https://github.com/lottie-react-native/lottie-react-native/actions/runs/7523023603/job/20485648831. @Victor0814gui

.github/workflows/ci-windows.yml
```diff
@@ -0,0 +1,48 @@
...
+        cache: 'yarn'
+
+    - name: Setup MSBuild
+      uses: microsoft/setup-msbuild@v1.1.3
+      with:
+        vs-version: '[17.0,)'
+        msbuild-architecture: x64
+
+    - name: Install node modules
+      run: yarn --pure-lockfile
```
TheRogue76, line 26: I believe the command should be `yarn install --pure-lockfile`, though i am not sure why you are trying to stop it from generating a lockfile, considering it is going to be thrown out as soon as the CI job is done

TheRogue76: We use `--frozen-lockfile` to pin the dependencies so that if some minor version is ever bumped, we don't start seeing unrelated issues

.github/workflows/ci-windows.yml
```diff
@@ -0,0 +1,48 @@
...
+        yarn build
+        yarn tsc
+
+    - name: Build x64 release
+      shell: powershell
+      run: yarn windows --release --no-packager --no-deploy --logging
+
+    # Workaround for a bug in package searching during deploy.
+    # The deploy script only searches windows/{*/bin/x64/Release,Release/*}, but the build step above placed the packages at windows/x64/Release.
+    # Copy the packages to Windows/Release before deploying.
```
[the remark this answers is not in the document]
Victor0814gui: I didn't pay attention to the comments; I have already removed them as they were not necessary.

.github/workflows/ci-windows.yml
Victor0814gui: I didn't pay attention to the comments; I have already removed them as they were not necessary.

Victor0814gui:
A basic summary of what I did in this last change: the dependency on ```react-native-windows``` in the example was at version ```0.71.18``` and has been updated to ```0.72.24```. Additionally, I modified the Example.sln file to use the correct paths for the node_modules folder, and updated the version of ```Microsoft.UI.Xaml```  from ```2.6.0``` to ```2.8.0```. Finally, concluding with the CI correction.

.github/workflows/windows-paper-build.yml
```diff
@@ -0,0 +1,34 @@
...
+          nuget-version: "5.x"
+
+      - name: Setup MSBuild
+        uses: microsoft/setup-msbuild@v1.1.3
+        with:
+          vs-version: "[17.0,)"
+          msbuild-architecture: x64
+
+      - name: Install node modules
+        run: yarn install
```
TheRogue76, line 30:
```suggestion
        run: yarn install --frozen-lock
```

TheRogue76:
> A basic summary of what I did in this last change: the dependency on `react-native-windows` in the example was at version `0.71.18` and has been updated to `0.72.24`. Additionally, I modified the Example.sln file to use the correct paths for the node_modules folder, and updated the version of `Microsoft.UI.Xaml` from `2.6.0` to `2.8.0`. Finally, concluding with the CI correction.

Left one change that i think is important. The rest looks good. I'll let the CI run and once all the stuff is resolved, we can merge. One question though, mostly out of curiosity, does the linking for `Example.sln` need to be handled manually? Does RNW not have some sort of Auto linking for these?

TheRogue76:
The CI failed again. Looks like it is missing an SDK:
Screenshot 2024-01-15 at 18 58 24

Victor0814gui:
I fixed the absence of the SDK and added an extra step to check which SDKs are installed👍.

TheRogue76:
The Workflow file is being reported as invalid. I believe you have a typo on Line 33:
https://github.com/lottie-react-native/lottie-react-native/actions/runs/7533290462

TheRogue76:
Hi.
Any progress / things we can help with?

TheRogue76:
Closing due to a lack of activity

PR #1160, opened 2024-01-16
feat: add experimental bridgeless support

TheRogue76:
I have spoken with a couple of Lib maintainers and read up a bit, and what i came up with is this:

1. ~~The `customDirectEventTypes` has existed since 0.67, which means it is ok to use~~ Looking a bit more closely, looks like it registers things in the global scope, and considering how the API is more or less internal, i am very hesitant about using it
2. We will mark this as experimental, so if we roll out a new version and it ends up causing issues, we can claw it back.
3. We will roll out a new version with this, give it some time, handle any sort of changes that the core team might do to this API, then once it is finalised, we will make the changes, handle all of our upcoming breaking changes (the Android font issue and the new install script from RN) and roll out a new V7 with Bridgelss

Testing on Android:
Screenshot 2024-01-16 at 20 43 04
Testing on iOS:
Screenshot 2024-01-16 at 21 01 15


TheRogue76:
I also love how iOS's event names are just straight up different. I will probably rename Android's event names so they are aligned with iOS's and i can remove the redundant ones

TheRogue76:
@matinzd Let me know what you think. I am not marking the PR as ready, since i also need to update fabric to 73.2, and i want your thoughts on this before i commit to all the change that would need

TheRogue76:
Found some interesting stuff. Apparantly we are using the wrong type in the codegen. We should be using `HostComponent` instead of `NativeComponentType` and we would be able to remove this direct access. I will continue working on this

TheRogue76:
Awaiting a response from Meta: https://github.com/facebook/react-native/issues/42477

TheRogue76:
This PR is now Unblocked by Meta. Once https://github.com/lottie-react-native/lottie-react-native/pull/1174 is merged i will update this PR again and merge

matinzd (approved):

matinzd (approved):

PR #1163, opened 2024-01-21
feat: add privacy manifest for iOS

TheRogue76:
This PR aims to make us compliant with Apple's upcoming privacy manifest requirements that will go into effect come Spring 2024. We will need to do another update once [Lottie iOS releases 4.4.0](https://github.com/airbnb/lottie-ios/pull/2252#issuecomment-1885925359) with their own privacy manifest as well, but for now, This is all that is required.
I also bumped our Lottie iOS dep to 4.3.4, just so everything is on the latest version

### DIFF
--- a/packages/core/ios/PrivacyInfo.xcprivacy
+++ b/packages/core/ios/PrivacyInfo.xcprivacy
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyTracking</key>
+	<false/>
+	<key>NSPrivacyTrackingDomains</key>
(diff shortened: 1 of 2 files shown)


packages/core/ios/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,14 @@
...
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyTracking</key>
+	<false/>
+	<key>NSPrivacyTrackingDomains</key>
+	<array/>
+	<key>NSPrivacyCollectedDataTypes</key>
+	<array/>
+	<key>NSPrivacyAccessedAPITypes</key>
```
TheRogue76, line 11: Lottie iOS embeds ZipFoundation, hence the need to declare `3B52.1`. We do not embed anything, so the declaration should be blank for us

matinzd (approved):

PR #1174, opened 2024-02-02
fix: update RN to fix the pipeline

TheRogue76:
The CI is broken because Cocoapods 1.15 came out and ruined everyone's day. This PR brings the CI back into working condition so i can wrap up Bridgeless


apps/fabric/ios/example.xcodeproj/project.pbxproj
TheRogue76, line 1: No idea why running Pod install with RN 73.3 added these, but it did. I double checked, and it is just manually added by it

matinzd (approved):

PR #1176, opened 2024-02-06
[Windows]Add windows to lottie published files

jfkm69:
## Motivation
Looks like when moving over to v6 of the Lottie package it appears that the files were copied over to a new folder destination and there were changes to the `package.json` to limit the range of published files, unfortunately, it looks like these changes excluded the windows folder. So while [the files are present in the repo](https://github.com/lottie-react-native/lottie-react-native/tree/master/packages/core/windows), they are not published as part of the package, thus react native windows developers are effectively locked out of consuming the newer versions of this package.

### Fixed
- Fixed missing windows directly in published files list in the package.json

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -14,7 +14,8 @@
     "android",
     "ios",
     "*.podspec",
-    "react-native.config.js"
+    "react-native.config.js",
+    "windows"
   ],
   "author": "Emilio Rodriguez <emiliorodriguez@gmail.com>",
   "homepage": "https://github.com/airbnb/lottie-react-native#readme",


chiaramooney (approved):

TheRogue76:
Hi @jfkm69 and @chiaramooney,
I am happy to support the Windows effort on LRN, but none of the current active maintainers have Windows devices to check if things are running correctly. I know there was an effort in https://github.com/lottie-react-native/lottie-react-native/pull/1158 to get things running again, but can you validate that this is all that is needed? I can merge it if you give a thumbs up on it

TheRogue76 (approved):

TheRogue76:
Your changes should now be available as of 6.7.0.

PR #1199, opened 2024-03-12
ADD UNSET when 0.72-rc2 doesn't have UNSET property

Cura-inthelab:


TheRogue76:
Hi @Cura-inthelab.
Let me come out and say thank you for the PR. I will go out on a limb and assume you meant 0.74-rc2 and not 0.72-rc2. We are aware of the change, and as of https://github.com/facebook/react-native/pull/43491 it has been reverted. The reason we haven't taken any actions for it is because the RCs are not finalized React Native versions, and are prone to change. I would recommend simply waiting for the stable release to come out. But super appreciate the initiative!

TheRogue76:
Closing. It is now part of 74 rc4: https://github.com/orgs/reactwg/projects/5/views/3?pane=issue&itemId=56772208

PR #1206, opened 2024-03-19
refactor(web): replace inline require with import

adamhari:
### Why

The inline require is no longer necessary now that the web implementation always uses `@dotlottie/react-player`.

Replacing it with an import simplifies typing in the file and surfaces dependency issues at compile-time.

### Testing Notes

- ran the `setup` and `paper:web` scripts after making the change and verified that the example app still works

image


TheRogue76 (approved):
Nice. Thank you for the contribution!🥳

PR #1209, opened 2024-03-29
chore: bump Lottie Android to 6.4.0

TheRogue76:
Been a while since we bumped up Android's Lottie, I see some new goodies in there, and we might as well keep it up to date.


TheRogue76:
Also, we should probably migrate to using toml. This is some ancient dark magic stuff

matinzd (approved):

PR #1248, opened 2024-06-28
refactor(web): replace deprecated @dotlottie/react-player with @lottiefiles/dotlottie-react

adamhari:
[`@dotlottie/react-player` has been deprecated](https://github.com/dotlottie/player-component?tab=readme-ov-file#%EF%B8%8F-this-package-is-no-longer-maintained-please-use-the-latest-dotlottie-web-here-for-the-latest-features-and-updates) in favor of [`@lottiefiles/dotlottie-react`](https://github.com/LottieFiles/dotlottie-web/tree/main/packages/react).

This PR refactors the web implementation to use the new player component.

I've found that the new web player provides for more control over layout and makes it easier to style things consistently with native because, [unlike the old web player](https://github.com/dotlottie/player-component/blob/master/packages/react-player/src/react-player.tsx#L306), [this one doesn't wrap the animation in a display: block `div` which cannot receive styling via props](https://github.com/LottieFiles/dotlottie-web/blob/main/packages/react/src/use-dotlottie.tsx#L27). 

# Testing

The example app has been modified to restore support for the web platform, which was removed as part of the `7.0.0` change set.

https://github.com/user-attachments/assets/28cec37e-ce55-43f8-9b94-3c0e4593ab9e


adamhari:
@arasrezaei I understand that you are responsible for maintaining the web platform, and I would greatly appreciate a review when you get a chance. Thank you!

TheRogue76:
@arasrezaei, If you can, please review this PR, i plan on merging this to develop if nothing else comes along.

packages/core/src/LottieView/index.web.tsx
```diff
@@ -24,151 +20,128 @@ const LottieView = forwardRef(
...
       progress,
+      onAnimationLoaded,
       onAnimationFailure,
       onAnimationFinish,
       onAnimationLoop,
     }: LottieViewProps,
-    ref: any
+    ref: any,
   ) => {
+    const [dotLottie, setDotLottie] = useState<DotLottie>(null);
```
TheRogue76, line 30: Shouldn't this be DotLottie | null?

packages/core/src/LottieView/index.web.tsx
```diff
@@ -24,151 +20,128 @@ const LottieView = forwardRef(
...
-    /**
-     *  If an error occured reset the key when the source changes to force a re-render.
-     */
     useEffect(() => {
-      if (isError) {
-        setKey((prevKey) => prevKey + 1);
-        setIsError(false);
-      }
-    }, [source]);
+      if (dotLottie) {
```
TheRogue76, line 37: Is this because dotLottie can be not installed, or is this a thing that this package needs by default?

TheRogue76: Read the documentation. Very strange way of adding event handlers but ok

packages/core/src/LottieView/index.web.tsx
adamhari: Yes, typescript won't complain in this project because `strict` is false in the tsconfig.

I made the necessary changes for the web implementation to be compliant with strict mode should we choose to enable it.

packages/core/src/LottieView/index.web.tsx
adamhari: It's because `dotLottie` can be null when this effect runs, particularly when the `sources` prop is not provided.

TheRogue76:
@adamhari We can merge this by next week so that @arasrezaei can have ample time to review it. But looks great so far to me.

packages/core/src/LottieView/index.web.tsx
adamhari: Yeah they force us to use the web api via the ref, would have been nice if the library handled that for us via props but alas. Still cleaner than the previous implementation!

arasrezaei:
that's cool, LGTM

TheRogue76 (approved):

TheRogue76:
Thank you both

PR #1253, opened 2024-07-05
fix: remove rn 72 hack  (PR: 3)

TheRogue76:
RN 72 is entering EOL, and now that Lottie-iOS requires iOS 13 and above, we can safely discard the old hack that we had here
https://github.com/reactwg/react-native-releases/blob/main/docs/support.md


matinzd:
Let's add a compatibility table to the readme later.

matinzd (approved):

PR #1277, opened 2024-09-02
Fix incorrect .lottie URI path check

geomaster:
The check for whether a URL looks like a `.lottie` file was over-zealous and would trigger if the URL had `.lottie` anywhere in it, regardless of context. Make this more precise by parsing the URL and checking only the extension of the filename part.

### DIFF
--- a/packages/core/src/LottieView/utils.ts
+++ b/packages/core/src/LottieView/utils.ts
@@ -1,5 +1,15 @@
 import { Image } from 'react-native';
 
+function looksLikeDotLottieURI(uri: string): boolean {
+  try {
+    const parsedURL = new URL(uri);
+    return parsedURL.pathname.endsWith('.lottie');
+  } catch (e) {
+    // Failed to parse URL, assume it's an ordinary JSON
+    return false;
+  }
+}
+
 function parsePossibleSources(source):
   | {
       sourceURL?: string;
@@ -19,8 +29,7 @@ function parsePossibleSources(source):
   }
 
   if (typeof source === 'object' && uri) {
-    // uri contains .lottie extension return sourceDotLottieURI
-    if (uri.includes('.lottie')) {
+    if (looksLikeDotLottieURI(uri)) {
       return { sourceDotLottieURI: uri };
     }
 


WrathChaos (approved):

TheRogue76 (approved):

TheRogue76:
Looks good. Good change @geomaster! I'll let the CI run and then merge

TheRogue76:
Reverting, as Hermes does not implement URL correctly at this moment, and needs a polly fill https://github.com/facebook/hermes/discussions/1072

geomaster:
@TheRogue76 Can you clarify your position on this functionality in light of https://github.com/lottie-react-native/lottie-react-native/pull/1277#issuecomment-2351108198, given that the PR was closed? Our users are unable to use lottie-react-native with Lotties that we host for them, since our domain (https://cdn.lottielab.com) contains the substring `.lottie`.

We would ideally like to contribute an acceptable fix to upstream, rather than advising users to use our fork of the library.

Thanks in advance!

TheRogue76:
> @TheRogue76 Can you clarify your position on this improvement in light of https://github.com/lottie-react-native/lottie-react-native/pull/1277#issuecomment-2351108198, given that the PR was closed? Our users are unable to use lottie-react-native with Lotties that we host for them, since our domain (https://cdn.lottielab.com) contains the substring `.lottie`.
> 
> 
> 
> We would ideally like to contribute an acceptable fix to upstream, rather than advising users to use our fork of the library.
> 
> 
> 
> Thanks in advance!

I am more than happy to accept any fixes upstream that fix things for all users, as you saw when i approved and merged the original PR.

In this case, this PR caused a hard crash in release mode for Android and iOS for bare RN, hence me reverting it.

Couple of stances we have:
- This library is supposed to work for both RN bare and Expo.
- Unless there are no other ways of doing it, we would rather not force people to install third party libraries for polly fill (we are having a lot of issue with Lottie Web, just because it is using a third party under the hood, and if i remember correctly that was also because another company wanted to use the library and promote it for web, and they have since abandoned maintenance on it).

I understand the frustration, and if you find another fix, i am more than happy to once again merge it

if you can do it without a pollyfill, great! If you need to use a pollyfill, at the very least refactor that section so that all the code uses it, not just one function

Issue #1286, opened 2024-09-21
Typing for progress does not match documentation


trajano:
### Description

The documentation states it allows Animated number but the typing only allows `number`

### Steps to reproduce

n/a

### Snack or a link to a repository

n/a

### Lottie React Native version

6.7.0

### React Native version

0.74.5

### Platforms

Android, iOS

### Workflow

Expo Dev Client

### Architecture

Fabric (New Architecture)

### Build type

Debug app & production bundle

### Device

Real device

### Acknowledgements

Yes

TheRogue76:
Hi @trajano 
Where in the documentation did you see that?
From ReadMe:
<img width="884" alt="Screenshot 2024-09-21 at 10 29 26" src="https://github.com/user-attachments/assets/30efaa44-3ad9-4d5f-bc2e-a6d5458454c3">
From the actual API md file:
<img width="1022" alt="Screenshot 2024-09-21 at 10 28 37" src="https://github.com/user-attachments/assets/f67a2570-df7b-471c-b3af-fbc2aa8b9017">
Neither part is claiming this

TheRogue76:
Please note that 
`const AnimatedLottieView = Animated.createAnimatedComponent(LottieView);`
Is what is taking the Animated progress, NOT Lottie View

trajano:
https://github.com/lottie-react-native/lottie-react-native/blob/b7b4f8b6ab8476478608f1fe51308eeeb06b883a/packages/core/src/types.ts#L48C1-L55C21

It may want to note that it requires it to be an `AnimatedLottieView` for it to take an Animated value.

LukasMod:
I had similar issue but with reanimated. It happened for me after upgrading a lot of libraries (including rn, reaniamted and lottie) so not sure what specifically caused it. I started getting type error when use react-native-reanimated to animate the progress without useAnimatedProps (before it was working fine).

```
import Lottie from 'lottie-react-native';
import Animated, { useSharedValue } from 'react-native-reanimated';

const AnimatedLottieView = Animated.createAnimatedComponent(Lottie);

const progress = useSharedValue<number>(0);

<AnimatedLottieView
    source={animation}
    progress={progress} // Type error here showed up
/>
```

type error log:

```
Type 'SharedValue<number>' is not assignable to type 'number | SharedValue<number | undefined> | undefined'.
  Type 'SharedValue<number>' is not assignable to type 'SharedValue<number | undefined>'.
[...]

PR #1288, opened 2024-09-21
Add note about using Animated

trajano:
Should be enough to address https://github.com/lottie-react-native/lottie-react-native/issues/1286

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -50,6 +50,9 @@ export interface LottieViewProps {
    * represents the normalized progress of the animation. If you update this prop, the
    * animation will correspondingly update to the frame at that progress value. This
    * prop is not required if you are using the imperative API.
+   *
+   * To use the `Animated` value, the component needs to be wrapped using
+   * `Animated.createAnimatedComponent(LottieView)`
    * @platform ios, android, windows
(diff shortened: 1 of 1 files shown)


packages/core/src/types.ts
```diff
@@ -50,6 +50,9 @@ export interface LottieViewProps {
    * represents the normalized progress of the animation. If you update this prop, the
```
TheRogue76, line 50: I'd rather just remove the reference to Animated number at the beginning of this instead of adding this clarification. The internals of `createAnimatedComponent` should not be something this library points to. This is just a remnant of V5's behaviour. The Sample in doc is enough.

packages/core/src/types.ts
```diff
@@ -50,6 +50,9 @@ export interface LottieViewProps {
...
    * animation will correspondingly update to the frame at that progress value. This
    * prop is not required if you are using the imperative API.
+   *
+   * To use the `Animated` value, the component needs to be wrapped using
+   * `Animated.createAnimatedComponent(LottieView)`
```
matinzd, line 55: This is not really needed as you can wrap any component to in order to make the props animated. It's not just specific to lottie's library.

PR #1297, opened 2024-10-28
fix: remove depracated property from AndroidManifest

TheRogue76:
Been a very long time coming. Now that we only support 73 and above, we can safely get rid of this one


TheRogue76:
@matinzd Can you please take a look?

matinzd (approved):

PR #1298, opened 2024-10-30
Fix: Ensure onAnimationFinish Callback Fires Correctly in Fabric

MaxToyberman:
Problem
In the Fabric architecture, updateProps calls each property’s setter individually, so the order of setters impacts behavior.

The issue occurs because setSpeed is called after setSourceJson (which initializes LottieAnimationView), affecting the onAnimationFinish callback:

In the new architecture: animationView?.play() is triggered immediately, but without a completion callback.
In the old architecture: playIfNeeded() is used, which includes a completion callback, ensuring onAnimationFinish is fired.
Solution Options
To ensure consistent behavior between the new and old architectures, two potential solutions were considered:

**_Option 1_**: Adjust the order of setter calls so the behaviour will be the same on both archs and make the onAnimationFinish callback in the new architecture.

**_Option 2_**: Provide a completion callback directly in setSpeed, ensuring onAnimationFinish fires reliably. (can be breaking change)
This PR implements Option 1 to maintain consistency in callback behavior across architectures.

### DIFF
--- a/packages/core/ios/Fabric/LottieAnimationViewComponentView.mm
+++ b/packages/core/ios/Fabric/LottieAnimationViewComponentView.mm
@@ -53,6 +53,18 @@ using namespace facebook::react;
         [_view setResizeMode:RCTNSStringFromString(newLottieProps.resizeMode)];
     }
 
+    if(oldLottieProps.progress != newLottieProps.progress) {
+        [_view setProgress:newLottieProps.progress];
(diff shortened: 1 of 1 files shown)


TheRogue76 (approved):

TheRogue76:
Tried it out on your repo, looks good. Amazing find!

PR #1304, opened 2024-11-21
fix: play lottie animation without restarting, similar to iOS

### LINKED ISSUE #1303: Playing lottie file with imperative Api restarts animation on Android
### Description

Hi,

When I try to play an animation on Android and it is halfway through playing, it restarts the animation.
On iOS, it just keeps on playing and that how I would intuitively think it would work, but that is not the case on Android:

| **iOS** | **Android** |
| ---------- | --------- |
| <video src="https://github.com/user-attachments/assets/5b2217f8-2205-4b3d-a3ec-dbe45d6b8e4d"></video> | <video src="https://github.com/user-attachments/assets/4a053000-e31b-4e32-8703-40a1cfa5a3c3"></video> |

### Steps to reproduce

1. Clone the [repo](https://github.com/itsramiel/AndroidLottie/tree/imperative-play) with branch `imperative-play`
2. run the app on iOS, Android
3. press on button multiple times and notice how android restarts

https://github.com/itsramiel/AndroidLottie/tree/imperative-play

7.1.0

### React Native version

0.75.4

### Platforms

Android, iOS

### Workflow

React Native

### Architecture

Paper (Old Architecture)

### Build type

Debug app & dev bundle

### Device

Android emulator

### Acknowledgements

Yes

itsramiel:
## Summary
Closes #1303 

Aligns Android's play behaviour with iOS'

### DIFF
--- a/packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManagerImpl.kt
+++ b/packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManagerImpl.kt
@@ -81,6 +81,7 @@ internal object LottieAnimationViewManagerImpl {
 
     @JvmStatic
     fun play(view: LottieAnimationView, startFrame: Int, endFrame: Int) {
+        val restartAnimation = startFrame != -1 && endFrame != -1
         Handler(Looper.getMainLooper()).post {
             if (startFrame != -1 && endFrame != -1) {
                 if (startFrame > endFrame) {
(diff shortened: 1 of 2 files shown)


TheRogue76:
This looks ok to me. @matinzd any thoughts on your side?

packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManagerImpl.kt
```diff
@@ -94,16 +95,31 @@ internal object LottieAnimationViewManagerImpl {
...
                 }
+            } else {
+                val actualStartFrame = view.composition?.startFrame?.toInt()
+                val actualEndFrame = view.composition?.endFrame?.toInt()
+
+                val minFrame = view.minFrame.toInt()
+                val maxFrame = view.maxFrame.toInt()
+                if (actualStartFrame != null && actualEndFrame != null && (minFrame != actualStartFrame || maxFrame != actualEndFrame)) {
+                    view.setMinAndMaxFrame(actualStartFrame, actualEndFrame)
+                }
```
matinzd, line 106: Can you explain what is happening here?

packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManagerImpl.kt
```diff
@@ -81,6 +81,7 @@ internal object LottieAnimationViewManagerImpl {
 
     @JvmStatic
     fun play(view: LottieAnimationView, startFrame: Int, endFrame: Int) {
+        val restartAnimation = startFrame != -1 && endFrame != -1
         Handler(Looper.getMainLooper()).post {
             if (startFrame != -1 && endFrame != -1) {
```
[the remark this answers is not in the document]
itsramiel: Done. I also renamed the variable since it is not exactly restarting the animation and didnt seem clear to me coming to the pr after some time. Hopefully you agree with the new name.

packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManagerImpl.kt
itsramiel: This makes sure that if a `play` methods is called without start/end, it really plays it from the original asset's start till its end. If this was not done and the user calls `play(<start-frame>, <end-frame>)` then wants to play the entire thing with `play`, then this second play would not play from the very beg till the very end. Instead it would play from the prev passed `play` call because of the `setMinAndMaxFrame`.

Here simply we are making sure the min and max is reset to the original if no start/end is passed.

matinzd: Thank you so much! However, I think we need to refactor this code because it can lead to race conditions. While this piece serves as a workaround for now, the root issue, in my opinion, is the race condition. Especially, on line 85, we are scheduling a task on the main thread, and calling `play` multiple times can result in conflicting configurations on the view.

matinzd (approved):

TheRogue76 (approved):

TheRogue76:
Awesome work @itsramiel ❤️

Issue #1098, opened 2023-08-28
dotLottie files are not working after codepush release


jameskenyon:
### Description

Lottie not loading on some Android devices when using `.lottie` files. 
We have had reports of the Lotties not displaying for users on some device models. Manufacturers of these devices include, but are not limited to: Samsung, Huawei, and Xiaomi. 
Tested on Google Pixel (could not reproduce) and Huawei (could reproduce).

We have reverted back to using `.json` files which has resolved the issues.

### Steps to Reproduce

Animated Lottie View: 
`
import Lottie, { LottieViewProps } from 'lottie-react-native'
import Animated, { Easing, useSharedValue, withTiming } from 'react-native-reanimated'

const AnimatedLottieView = Animated.createAnimatedComponent(Lottie)
`

Lottie component configuration: 
`
<AnimatedLottieView
	source={require(path_to_lottie)}
        autoPlay
	loop
/>
`

**Expected behavior:** 

![Screenshot 2023-08-28 at 15 13 50](https://github.com/lottie-react-native/lottie-react-native/assets/66872664/bfb6b6df-83c2-4809-8a60-de5f2a119dbf)

**Actual behavior:** 

<img width="299" alt="Screenshot 2023-08-28 at 15 14 47" src="https://github.com/lottie-react-native/lottie-react-native/assets/66872664/22bfcb99-95e5-420e-9e3f-c5bacc3ac5de">

### React Native Environment

Node: 16.18.1
Yarn: 1.22.19 
npm: 8.19.2 
Watchman: 2023.08.14.00 
CocoaPods: 1.12.1
react: 18.2.0 => 18.2.0 
react-native: 0.71.12 => 0.71.12 

### Lottie Version

Version: 6.2.0

walterholohan:
Hey @matinzd this is a continuation of the bug https://github.com/lottie-react-native/lottie-react-native/issues/1090. @jameskenyon is part of our team. 

Let me know if you need any help to debug further.

Thanks

ifsnow:
I think since v6 you have to explicitly specify the size of the animation with a style. (or flex: 1)

```
<AnimatedLottieView
  source={require(path_to_lottie)}
  autoPlay
  loop 
  style={{
    width: xxx,
    height: xxx,
  }}
/>
```

In my case, if I don't specify a size, it doesn't show up.

matinzd:
Hi,

[...]

PR #1319, opened 2025-01-13
feat(android): support locally stored dotlottie assets  

swizes:
When the application uses OTA to deliver updates, the dotlottie assets can't be loaded on android.

Let's say that a dotlottie asset is located in the src/assets/animations folder.

Scenario 1 - No OTA update is installed on the application
When we are passing this file as source to the LottieView, Android layer of this lib uses following uri to set it as dotLottieSourceUri `src_assets_animations_myanimation`. Lottie can load this file as expected.

Scenario 2 - A irrelevant OTA update is pushed and installed on the device

The same file, which we used above is now passed to the android layer with following uri

`file:///data/user/0/com.example.app/files/output/raw/src_assets_animations_myanimation.lottie`

Unfortunately the current logic can't handle any URI which has file:// protocol in it. Therefore users are not able to see the dotlottie animations in the app anymore.

To be able solve this issue, I have improved the `commitChanges` logic to handle such cases.

It will solve https://github.com/lottie-react-native/lottie-react-native/issues/1098

### DIFF
--- a/packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
+++ b/packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
@@ -3,6 +3,7 @@ package com.airbnb.android.react.lottie
 import android.graphics.ColorFilter
 import android.graphics.Typeface
 import android.net.Uri
+import android.util.Log
 import android.widget.ImageView
 import com.airbnb.lottie.LottieAnimationView
 import com.airbnb.lottie.LottieDrawable
@@ -36,6 +37,8 @@ import java.io.FileInputStream
  */
 class LottieAnimationViewPropertyManager(view: LottieAnimationView) {
     private val viewWeakReference: WeakReference<LottieAnimationView>
+    private val TAG = "lottie-react-native"
+
 
     /**
      * Should be set to true if one of the animationName related parameters has changed as a result
@@ -72,7 +75,7 @@ class LottieAnimationViewPropertyManager(view: LottieAnimationView) {
                 return ReactFontManager.getInstance()
                     .getTypeface(fontFamily, UNSET, UNSET, view.context.assets)
             }
-        
+
             override fun fetchFont(fontFamily: String, fontStyle: String, fontName: String): Typeface {
                 val weight = when (fontStyle) {
                     "Thin" -> 100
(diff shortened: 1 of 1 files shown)


packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
```diff
@@ -129,7 +129,18 @@ class LottieAnimationViewPropertyManager(view: LottieAnimationView) {
...
-            var file = File(assetName)
+
+            var file: File? = null
+
+            val scheme = runCatching { Uri.parse(assetName).scheme }.getOrNull()
+            // if the asset path has file:// prefix, which indicates locally stored file, parse the path to be able to load it properly
+            // This is useful for apps, which are using OTA (CodePush, Expo-Updates etc.)
+            if (scheme == "file") {
+                val uri = Uri.parse(assetName)
+                file = File(uri.path!!)
```
TheRogue76, line 140: Can you do this without force unwraping? File is already null, you can just do `uri.path?.let { }`.

TheRogue76:
Very fun PR, left a comment, also, does this work on iOS by default with OTA?

TheRogue76:
CC: @matinzd . I know you are not a fan of hacks but i think this one could be pretty valid

packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
TheRogue76: And maybe log into Logcat something if File is still null after all the operations (uri.path somehow did not resolve to something)

swizes:
> t, also, does this work on iOS by default wi

It works by default on iOS

packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
matinzd: Same comment. Please fix this and it's good to go.

TheRogue76 (approved):

packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
TheRogue76: Beautiful, approved and set the CI to run, can merge after it gives us the go ahead

matinzd (approved):

swizes:
@matinzd @TheRogue76 when can we expect the patch version to be released?

matinzd:
Released! https://github.com/lottie-react-native/lottie-react-native/releases/tag/v7.2.2

PR #1324, opened 2025-02-04
Depend on lottie-ios ~> 4.5.0 instead of fixed 4.5.0

e-sung:
I have a lottie file that has performance issue with lottie-ios 4.5.0, but that issue is resolved with 4.5.1 
Yet, for now I can't depend on lottie-ios 4.5.1 because lottie-react-native is depending on lottie-ios 4.5.0 fixed version.

If it's not intended, as [cocoapods guide recommends ](https://guides.cocoapods.org/syntax/podspec.html#dependency), I think it's good idea to allow minor version updates in lottie-ios

### DIFF
--- a/packages/core/lottie-react-native.podspec
+++ b/packages/core/lottie-react-native.podspec
@@ -22,7 +22,7 @@ Pod::Spec.new do |s|
     'Lottie_React_Native_Privacy' => ['ios/PrivacyInfo.xcprivacy'],
   }
 
-  s.dependency 'lottie-ios', '4.5.0'
+  s.dependency 'lottie-ios', '~> 4.5.0'
 
   s.swift_version = '5.9'
 


TheRogue76:
Hi @e-sung 
It is intended as we had one patch version of Lottie at one point that bumped the iOS minimum requirements, and that ended up breaking things for people since their RN iOS dep was way lower. LRN is a wrapper for Lottie-iOS, and in library land, it is best to be as specific as possible. The cocoapods recommendation only really applies to app dependencies, not libraries.

Regarding your issue, I think if you do the bump to 4.5.1 in pod spec, we should be good to do a little patch release. So feel free to update your PR and i can have it merged.

e-sung:
sorry I missed your kind comment. 
I think our project can use my forked repo for a while. I'll make a PR when really needed. thank you always!

PR #1336, opened 2025-04-05
Lottie V8, now with Nitro

TheRogue76:
It has been entirely too painful of an experience trying to maintain this package. It is not easy to reason about, meaning that unless me or @matinzd actually put everything aside, usually this package is not really getting the love it deserves from other parties.

Most of this comes down to two simple things: ObjC and the old architecture. ObjC is nobody's cup of tea, and isn't even what Lottie iOS uses either, it is just a wrapper that lets the actual Swift API of Lottie to talk with the actual C++ API of RN. It's a stop gap mess.
The old architecture is the second part of this mess. It has forced us to have two implementations of every API that Android has, and then remap them all back into a singular `LottieAnimationViewManagerImpl`. Its a mess, it's ugly to look at, and not what i want this to be. I want a singular clean interface, built for the modern age.

What i want for Lottie is a simple abstraction, close to hardware, close to what it really is, in modern languages that everyone can use. The goals are as follows:

1. Modern languages: Everyone should be able to contribute to this package, be it a native iOS dev who has worked with Lottie iOS, a Native android dev who has worked with Lottie Android, or some JS/TS dev who knows a decent amount about JS/TS
2. The same exact API as V7. Wanna upgrade to V8? Make sure your RN version is high enough, and then just bump Lottie RN and you are done. No breaking API changes.
3. New arch only. I do not want to carry over the bridge. It has been annoying keeping two of these maintained for years. One architecture, THE architecture of RN.

So we are going with Nitro. It will require some pretty high minimum RN version, but it will be worth it. We can make a V7 branch of the current Lottie, and put all the fixes in there so people can keep using the old implementation for a while longer.


TheRogue76:
Base View is now loading fine:
Screenshot 2025-04-05 at 16 55 01
Android is still not picking up the view though, so need to investigate

TheRogue76:
Alright, Android fixed:
Screenshot 2025-04-05 at 17 59 16

TheRogue76:
Whelp, guess i am committed now:

https://github.com/user-attachments/assets/e0799fd6-1c9c-4e64-be54-9d57a72e94cd

CODEOWNERS
matinzd, line 1: We can add our usernames.

TheRogue76:
iOS working now
Screenshot 2025-04-12 at 18 18 23

TheRogue76:
Color filters working just fine on iOS now!
Screenshot 2025-04-12 at 19 28 05

TheRogue76:
iOS's commands are still buggy. To figure out what is exactly wrong with them. Afterwards we can do a little RC or Alpha release

TheRogue76:
@matinzd

mrousavy:
> iOS's commands are still buggy.

Have you explored using [Nitro Views](https://nitro.margelo.com/docs/view-components) instead of normal Fabric views? Then you're using Nitro for everything, even callbacks. No more commands, no ObjC

Issue #1350, opened 2025-06-16
[SOLVED] lottie-react-native 7.2.2 Build Error with React Native 0.80 - Kotlin Type Mismatch


Developer-Nijat:
## Problem

When building a React Native 0.80 project with `lottie-react-native` version 7.2.2, the build fails with the following Kotlin compilation errors:

```
> Task :lottie-react-native:compileDebugKotlin FAILED

e: file:///path/to/node_modules/lottie-react-native/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt:249:26 
Initializer type mismatch: expected 'Int', actual 'Int?'.

e: file:///path/to/node_modules/lottie-react-native/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt:250:13 
Type mismatch: inferred type is 'Int?', but 'Int' was expected.
```

## Environment
- React Native: 0.80
- lottie-react-native: 7.2.2
- Platform: Android
- Build tool: Gradle/Kotlin

## Root Cause

The issue occurs in the `parseColorFilter` method of `LottieAnimationViewPropertyManager.kt`. The problem is that `ColorPropConverter.getColor()` now returns a nullable `Int?` type, but the code expects a non-nullable `Int`.

This is due to React Native 0.80's updated APIs becoming more null-safe, while the lottie-react-native library hasn't been updated to handle the new type signatures.

## Quick Fix

Navigate to the problematic file:
```
node_modules/lottie-react-native/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
```

Find the `parseColorFilter` method (around line 245-250) and change this line:

**Before:**
```kotlin
val color: Int = if (colorFilter.getType("color") == ReadableType.Map) {
    ColorPropConverter.getColor(colorFilter.getMap("color"), view.context)
} else {
    colorFilter.getInt("color")
}
```

**After:**
```kotlin
val color: Int = if (colorFilter.getType("color") == ReadableType.Map) {
    ColorPropConverter.getColor(colorFilter.getMap("color"), view.context) ?: 0
} else {
    colorFilter.getInt("color")
}
```

The key change is adding `?: 0` to provide a default value (transparent) when the color conversion returns null.

### Snack or a link to a repository

[...]

PR #1351, opened 2025-06-17
Build Error with React Native 0.80 - Kotlin Type Mismatch

dishantwalia:
When building a React Native 0.80 project with lottie-react-native version 7.2.2, the build fails with the following Kotlin compilation errors:

> > Task :lottie-react-native:compileDebugKotlin FAILED
> 
> e: file:///path/to/node_modules/lottie-react-native/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt:249:26 
> Initializer type mismatch: expected 'Int', actual 'Int?'.
> 
> e: file:///path/to/node_modules/lottie-react-native/android/src/main/java/com/airbnb/android/reaThis pull request includes a small change to the `LottieAnimationViewPropertyManager` class in the `LottieAnimationViewPropertyManager.kt` file. The change ensures that the `getColor` method returns a default value of `0` if the result is null.
> [Copilot is generating a summary...]ct/lottie/LottieAnimationViewPropertyManager.kt:250:13 
> Type mismatch: inferred type is 'Int?', but 'Int' was expected.

Solution: The key change is adding ?: 0 to provide a default value (transparent) when the color conversion returns null.

### DIFF
--- a/packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
+++ b/packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
@@ -247,7 +247,7 @@ class LottieAnimationViewPropertyManager(view: LottieAnimationView) {
         view: LottieAnimationView
     ) {
         val color: Int = if (colorFilter.getType("color") == ReadableType.Map) {
-            ColorPropConverter.getColor(colorFilter.getMap("color"), view.context)
+            ColorPropConverter.getColor(colorFilter.getMap("color"), view.context) ?: 0
         } else {
             colorFilter.getInt("color")
         }


dishantwalia:
https://github.com/lottie-react-native/lottie-react-native/issues/1350

itsramiel:
> Solution: The key change is adding ?: 0 to provide a default value (transparent) when the color conversion returns null.
Does 0 really return a transparent color? I thought it returns an opaque black

packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
```diff
@@ -247,7 +247,7 @@ class LottieAnimationViewPropertyManager(view: LottieAnimationView) {
         view: LottieAnimationView
     ) {
         val color: Int = if (colorFilter.getType("color") == ReadableType.Map) {
-            ColorPropConverter.getColor(colorFilter.getMap("color"), view.context)
+            ColorPropConverter.getColor(colorFilter.getMap("color"), view.context) ?: 0
```
itsramiel, line 250: 0 is quite vague. using Colors is clear imo. Maybe Colors.Black?

```suggestion
            ColorPropConverter.getColor(colorFilter.getMap("color"), view.context) ?: Colors.Black
```

itsramiel:
@matinzd Can we get support here?

packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
itsramiel: or `Color.TRANSPARENT` if transparent is wanted. I just think using the `Color` api has clearer intent

TheRogue76:
I'll set the CI to run, if it is all good, then we can merge and give out a build

PR #1359, opened 2025-07-02
fix(ios): wrong animation when switching between raw json and string source in fabric

### LINKED ISSUE #1358: Incorrect animation displayed on iOS in new architecture when switching between json animations and local assets
### Description

Fixes an issue where a lottie view switching between a raw json (imported on the JS side) and a local asset (added to the iOS project) leads to the wrong animation being presented. The issue is only happening in the new architecture and on iOS only (see video).

Expected:
https://github.com/user-attachments/assets/5e2b4eb8-4098-4c38-b98f-f268cf167d28

Actual:
https://github.com/user-attachments/assets/28391ef0-59f4-4bb9-9c93-124303546262

### Steps to reproduce

Run the repro app and press the toggle button. The string `Should show watermelon` reflects the correct state but the lottie view shows the wrong animation after the third press.

https://github.com/ismarbesic/lottie-react-native-bug-repro

7.2.3

### React Native version

0.80.0

### Platforms

iOS

### Workflow

React Native

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Acknowledgements

Yes

ismarbesic:
Fixes an issue where a lottie view switching between a raw json (imported on the JS side) and a local asset (added to the iOS project) leads to the wrong animation being presented. The issue is only happening in the new architecture and on iOS only.

Upon inspection, when updateProps() is called, the local properties `sourceJson` and `sourceName` will not be updated properly and therefore when switching between the two formats, the old value in "sourceName" will still be set when a json animation is presented. When switching back, the check `newSourceName == sourceName` will equal to true and therefore the new animation won't be presented as it should.

The local properties don't seem to be used anywhere and can just create confusion/incorrect state when switching between different sources. To mitigate this issue, the properties have been removed which means that whenever the source changes, the animation will be changed as well as long as the string is valid. In the old architecture, the properties are always reset when the animation changes and therefore the issue isn't present there.

Tested on both the reproducer app and a custom app running the new architecture [here](https://github.com/ismarbesic/lottie-react-native-bug-repro).

Fixes #1358.

### DIFF
--- a/packages/core/ios/LottieReactNative/ContainerView.swift
+++ b/packages/core/ios/LottieReactNative/ContainerView.swift
@@ -17,9 +17,7 @@ class ContainerView: RCTView {
     private var progress: CGFloat = 0.0
     private var autoPlay: Bool = false
     private var loop: LottieLoopMode = .playOnce
-    private var sourceJson: String = ""
     private var resizeMode: String = ""
-    private var sourceName: String = ""
     private var colorFilters: [NSDictionary] = []
     private var textFilters: [NSDictionary] = []
     private var renderMode: RenderingEngineOption = .automatic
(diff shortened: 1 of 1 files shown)


matinzd:
Good job and thanks for the PR! This will be available in the next release.

PR #1368, opened 2025-08-04
fix: use default style props for web

### LINKED ISSUE #1366: [Web] Style should be applied by default
### Description

Hi, Thank you for this amazing library, I recently noticed you support web.
By replacing an old lottie library compatible with web on my `react-native-project`, I realized the lotties were huge and the property `style` didn't do anything on it

I discovered later that the web style should be passed by the `webStyle` props

I would like that `style` to be applied by default, when no `webStyle` is passed as props.

In this way, the Lottie component would share a same interface between each different platform, and `webStyle` would override `style` when needed

### Steps to reproduce

Try to modify style for a web Lottie component and see that it has no effect
Add a webStyle and see that it works

https://github.com/pass-culture/pass-culture-app-native/pull/8499

7.2.5

### React Native version

0.77

### Platforms

Web

### Workflow

React Native

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

None

### Acknowledgements

Yes

xlecunff-pass:
closes #1366

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -33,6 +33,15 @@ type TextFilterAndroid = {
   replace: string;
 };
 
+type CommonStyleKeys = Extract<keyof CSSProperties, keyof ViewStyle>;
+
+/**
+ * Unified style type for web and native
+ */
+type UnifiedViewStyle = {
+  [K in CommonStyleKeys]: CSSProperties[K] & ViewStyle[K];
+};
+
 /**
  * Properties of the LottieView component
  */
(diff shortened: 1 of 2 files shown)


packages/core/src/types.ts
```diff
@@ -33,6 +33,15 @@ type TextFilterAndroid = {
...
 
+type CommonStyleKeys = Extract<keyof CSSProperties, keyof ViewStyle>;
+
+/**
+ * Unified style type for web and native
+ */
+type UnifiedViewStyle = {
+  [K in CommonStyleKeys]: CSSProperties[K] & ViewStyle[K];
+};
+
```
matinzd, line 44: I don't think it's a good idea to have a unified style since css properties will not be applied to native code.

matinzd: Usually, people would think that if they apply those CSS properties, they should be applied to native apps as well. That can create confusion.

PR #1372, opened 2025-08-07
fix: bring backward compatibility for containerStyle prop

sharifhh:
the latest changes are breaking changes because now you are wrapping lottie view with a view

this change make the changes non-breaking for people who didn't use containerProps, so for those people if they was using LottieView normally without containerProps it will keep working as before.


matinzd (approved):
Thanks 🚀

PR #1399, opened 2026-02-20
[iOS][Fabric] Fix autoPlay, onAnimationLoaded and view-recycling bugs in New Architecture

wanderbon:
## Summary

This PR fixes five interconnected bugs affecting `LottieView` on iOS with the New Architecture (Fabric) when using `autoPlay={false}` and/or `loop={false}`. The bugs cause animations to start automatically despite `autoPlay={false}`, prevent `onAnimationLoaded` from firing, and break subsequent visits to screens containing a `LottieView`.

---

### Bug 1 — `setSpeed` unconditionally starts playback regardless of `autoPlay`

**File:** `ios/LottieReactNative/ContainerView.swift`

#### Root cause

`setSpeed` called `animationView?.play()` whenever speed was non-zero and the animation was not already playing, ignoring the `autoPlay` flag entirely:

```
// Before
if newSpeed != 0.0 {
    animationView?.animationSpeed = newSpeed
    if !(animationView?.isAnimationPlaying ?? true) {
        animationView?.play()   // fires unconditionally
    }
}
```

On every mount, LottieView's JS `defaultProps.speed = 1` causes `setSpeed(1)` to be called. This started the animation immediately, even with `autoPlay={false}`.

#### Fix

Remove the `play()` call from `setSpeed`. Playback responsibility belongs exclusively to `setAutoPlay` and `replaceAnimationView` → `playIfNeeded()`.

```
// After
if newSpeed != 0.0 {
    animationView?.animationSpeed = newSpeed
} else if animationView?.isAnimationPlaying ?? false {
    animationView?.pause()
}
```

### Bug 2 — Fabric view recycling: `autoPlay` not yet updated when `setSpeed` fires

**File**: `ios/Fabric/LottieAnimationViewComponentView.mm`

#### Root cause

In `updateProps:oldProps:`, `setSpeed` was called on line 64, while `setAutoPlay` was called on line 96. When Fabric recycles a native view previously used by a component with `autoPlay=true` (e.g. an animation that played once), the recycled `ContainerView._autoPlay` is still true at the moment `setSpeed` is called with the new component's speed. 
Even with Bug 1 fixed, any future code in `setSpeed` that checks `autoPlay` would see the stale value.
Additionally, loop had the same ordering problem.

#### Fix

Move `setAutoPlay` and `setLoop` to the top of `updateProps:oldProps:`, before `setSpeed` and all source setters:

```
// Now called first — before setSpeed and source setters
if (oldLottieProps.autoPlay != newLottieProps.autoPlay) {
    [_view setAutoPlay:newLottieProps.autoPlay];
}
if (oldLottieProps.loop != newLottieProps.loop) {
    [_view setLoop:newLottieProps.loop];
}
// ... resizeMode, progress, speed, source, etc.
```

### Bug 3 — `setSourceDotLottieURI` loses the loader view via ARC
**File**: `ios/LottieReactNative/ContainerView.swift`

#### Root cause

```
// Before
_ = LottieAnimationView(
    dotLottieUrl: url,
    configuration: lottieConfiguration,
    completion: { ... }
)
```

Assigning the temporary `LottieAnimationView` to `_` discards the only strong reference. ARC can deallocate the view before the async load completes, silently dropping the completion callback. As a result, `replaceAnimationView` is never called and `onAnimationLoaded` never fires.

#### Fix
Retain the loader in a stored property and release it inside the completion:

```
let loader = LottieAnimationView(
    dotLottieUrl: url,
    configuration: lottieConfiguration,
    completion: { [weak self] view, error in
        guard let self = self else { return }
        self.loadingAnimationView = nil  // release
        // ...
        self.replaceAnimationView(next: view)
    }
)
loadingAnimationView = loader
```

### Bug 4 — `animationLoaded` closure assigned after the event already fired

**File**: `ios/LottieReactNative/ContainerView.swift`

#### Root cause

In `replaceAnimationView`, `animationView.animationLoaded` was assigned after `addSubview`. For animations loaded via `dotLottieUrl`, the `LottieAnimationView` passed to the completion callback already has the composition loaded. Lottie's `animationLoaded` property uses a Swift `didSet` observer that immediately calls the closure if `animation != nil`. By assigning after `addSubview`, this immediate call was missed.

```
// Before (animationLoaded set last — too late for already-loaded views)
addSubview(next)
applyContentMode()
applyColorProperties()
playIfNeeded()
animationView?.animationLoaded = { ... }  // ← didSet fires immediately but nobody is listening yet
```

#### Fix

Assign `animationLoaded` first, before any other setup:

```
// After
animationView?.animationLoaded = { [weak self] _, _ in
    guard let self = self else { return }
    self.loadedCallback()
}
// then: backgroundBehavior, animationSpeed, addSubview, playIfNeeded...
```

### Bug 5 — `onAnimationLoaded` dropped when Fabric attaches event emitter after load completes (race condition + view pool recycling)

**Files**: `ios/Fabric/LottieAnimationViewComponentView.mm`, `ios/Fabric/LottieContainerView.h`, `ios/LottieReactNative/ContainerView.swift`

#### Root cause — race condition

In Fabric, `updateProps:oldProps:` is called before `updateEventEmitter:`. For local `file:// .lottie` assets, the async load in `setSourceDotLottieURI` completes very quickly (synchronously in some cases), meaning `onAnimationLoaded` fires while `_eventEmitter` is still `nil`. The existing `guard if (!_eventEmitter) return;` silently drops the event.

#### Root cause — view pool recycling

When a user navigates away from a screen (unmounting it) and then navigates back, Fabric recycles native views. The recycled `LottieContainerView` already has the animation loaded from the previous mount. Since `sourceDotLottieURI` hasn't changed, Fabric's prop diffing skips `setSourceDotLottieURI`, `replaceAnimationView` is never called, and `onAnimationLoaded` never fires for the new component instance.

#### Fix

Track whether the animation has been loaded in a new `isAnimationLoaded: Bool` property on ContainerView. Override `updateEventEmitter:` in the Fabric component view to emit `onAnimationLoaded` when `_view.isAnimationLoaded` is `true`. This method is called on mount (including recycled views) but not on re-renders, making it the correct place for this check.

```
// ContainerView.swift
@objc private(set) var isAnimationLoaded: Bool = false

// Set to true in loadedCallback(), reset to false in removeCurrentAnimationView()
```

```
// LottieAnimationViewComponentView.mm
- (void)updateEventEmitter:(const EventEmitter::Shared &)eventEmitter {
    [super updateEventEmitter:eventEmitter];
    // Covers both the race condition (load before emitter) and view pool recycling
    // (same source → setSourceDotLottieURI skipped → onAnimationLoaded never called)
    if (_view.isAnimationLoaded) {
        [self onAnimationLoaded];
    }
}
```

```
// LottieContainerView.h
@property (nonatomic, readonly) BOOL isAnimationLoaded;
```

### Affected platforms

iOS only (New Architecture / Fabric). Android and Old Architecture are not affected.

### Test scenario

1. Screen A contains a `LottieView` with `autoPlay={true}`.
2. Navigating from Screen A to Screen B, which contains a `LottieView` with `autoPlay={false}` and `loop={false}`.
3. Screen B animation must not auto-start on arrival.
4. Screen B animation must fire `onAnimationLoaded`.
5. Navigating away from Screen B and back must still fire `onAnimationLoaded` on return.

### DIFF
--- a/packages/core/ios/Fabric/LottieAnimationViewComponentView.mm
+++ b/packages/core/ios/Fabric/LottieAnimationViewComponentView.mm
@@ -49,16 +49,24 @@ using namespace facebook::react;
     const auto &oldLottieProps = *std::static_pointer_cast<const LottieAnimationViewProps>(_props);
     const auto &newLottieProps = *std::static_pointer_cast<const LottieAnimationViewProps>(props);
 
+    // autoPlay and loop must be applied before speed and source so that playback-triggering
+    // setters (setSpeed, replaceAnimationView) always see the final autoPlay value.
+    // This is critical when Fabric recycles a native view whose previous autoPlay=true is
+    // still set at the moment setSpeed fires.
+    if(oldLottieProps.autoPlay != newLottieProps.autoPlay) {
+        [_view setAutoPlay:newLottieProps.autoPlay];
+    }
+
+    if(oldLottieProps.loop != newLottieProps.loop) {
+        [_view setLoop:newLottieProps.loop];
+    }
+
     if(oldLottieProps.resizeMode != newLottieProps.resizeMode) {
         [_view setResizeMode:RCTNSStringFromString(newLottieProps.resizeMode)];
     }
 
     if(oldLottieProps.progress != newLottieProps.progress) {
         [_view setProgress:newLottieProps.progress];
-    }
-
-    if(oldLottieProps.loop != newLottieProps.loop) {
-        [_view setLoop:newLottieProps.loop];
     }
 
     if(oldLottieProps.speed != newLottieProps.speed) {
@@ -93,11 +101,24 @@ using namespace facebook::react;
         [_view setRenderMode:RCTNSStringFromString(newLottieProps.renderMode)];
     }
 
-    if(oldLottieProps.autoPlay != newLottieProps.autoPlay) {
-        [_view setAutoPlay:newLottieProps.autoPlay];
-    }
-
     [super updateProps:props oldProps:oldProps];
+}
+
+- (void)updateEventEmitter:(const EventEmitter::Shared &)eventEmitter
+{
+    [super updateEventEmitter:eventEmitter];
+
+    // Emit onAnimationLoaded if the animation is already loaded in the native view.
+    // This covers two cases:
+    // 1. Fabric view pool recycling: same sourceDotLottieURI → Fabric skips setSourceDotLottieURI
+    //    so the native side never calls loadedCallback again, but the animation IS loaded.
+    // 2. Race condition: async load completed before _eventEmitter was attached (updateProps
+    //    fires before updateEventEmitter: in Fabric commit order).
+    // updateEventEmitter: is called on mount (including recycled views) but NOT on re-renders,
+    // so this emit is safe and won't duplicate events during normal prop updates.
+    if (_view.isAnimationLoaded) {
+        [self onAnimationLoaded];
+    }
 }
 
 #pragma mark - Native Commands
@@ -160,6 +181,8 @@ using namespace facebook::react;
 - (void)onAnimationLoaded
 {
     if(!_eventEmitter) {
+        // _eventEmitter not yet attached; updateEventEmitter: will emit once it arrives
+        // (it checks _view.isAnimationLoaded which is already true at this point).
         return;
     }
 
--- a/packages/core/ios/Fabric/LottieContainerView.h
+++ b/packages/core/ios/Fabric/LottieContainerView.h
@@ -13,6 +13,7 @@
 
 @interface LottieContainerView : RCTView
 @property (nonatomic, weak) id <LottieContainerViewDelegate> _Nullable delegate;
+@property (nonatomic, readonly) BOOL isAnimationLoaded;
 - (void)traitCollectionDidChange:(UITraitCollection * _Nullable)previousTraitCollection;
 - (void)setSpeed:(CGFloat)newSpeed;
 - (void)setProgress:(CGFloat)newProgress;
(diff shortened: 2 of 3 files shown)


benjaminharringtonrose:
`removeCurrentAnimationView()` resets `isAnimationLoaded` and clears `animationView`, but does not nil out `loadingAnimationView`. If a dotLottie URL is loading asynchronously and a different source type (e.g., via `setSourceURL` or `setSourceJSON`) replaces the animation, the stale dotLottie completion handler will still fire, call `replaceAnimationView(next:)`, and overwrite the correct animation with the old one. `loadingAnimationView` needs to be cleared in `removeCurrentAnimationView()` so ARC can deallocate the in-flight loader and prevent its completion from executing - does that sound right @wanderbon ?

nicobermudez (approved):

wanderbon:
> `removeCurrentAnimationView()` resets `isAnimationLoaded` and clears `animationView`, but does not nil out `loadingAnimationView`. If a dotLottie URL is loading asynchronously and a different source type (e.g., via `setSourceURL` or `setSourceJSON`) replaces the animation, the stale dotLottie completion handler will still fire, call `replaceAnimationView(next:)`, and overwrite the correct animation with the old one. `loadingAnimationView` needs to be cleared in `removeCurrentAnimationView()` so ARC can deallocate the in-flight loader and prevent its completion from executing - does that sound right @wanderbon ?

Yes, thanks, that makes sense. Without clearing loadingAnimationView, an in-flight async load can still complete and override the currently set animation. Resetting it in removeCurrentAnimationView() would prevent stale completions from mutating state and let ARC clean up the loader. I’ve made the changes accordingly.

adids1221:
Hey, just wanted to add that we're hitting the bug this PR fixes in production.
Lottie-based loader invisible on iOS cold start with New Arch.
would love to see this merged. 
please update if there is any progress with this pr.

samyos (approved):

s-ol:
We're also  blocked on a release here.

We need to update lottie to work with recent expo versions, which is mandatory to migrate to 16kb page sizes on android. But upgrading lottie breaks our app on iOS.

s-ol:
@matinzd have you had a chance to take a look at this?
We're considering doing a build with this branch to verify if that solves our use-case.

nicobermudez:
any update on this?

s-ol:
not an issue, not stale

PR #1405, opened 2026-03-16
Fix bug we observed using Lottie with Fabric architecture view recycling

fadookie:
Intermittently, contents of a `LottieView` would be rendered in other completely different places in the UI although this seems to have been isolated to other, unrelated `LottieView`s. After investigating, we believe this bug started occuring in our codebase after enabling Fabric architecture.

Looking at how expo's `ExpoFabricView` handles this, they completely disable view recycling: https://github.com/expo/expo/blob/ace67ea3ac7d96d6b44f5ec2167dff4f75e14313/packages/expo-modules-core/ios/Fabric/ExpoFabricView.swift#L160-L164

In our testing, this fixes the bug and seems safe to do as it's just an optimization. In the future if more thorough testing around properly handling reinitialzing a recycled view is done, this strategy can be changed, but completely disabling it seemed the safest strategy for now.


SYoder1:
I am seeing the same issue in my app. I can't reproduce it, but other devs can. iOS, lottie-react-native 7.3.5, RN 0.81, new arch.

benjaminharringtonrose:
I was hoping to see this PR - I was experiencing the same issue. Tested with iPhone 17 iOS 26.4, RN 0.81, lottie-react-native 7.3.5

matinzd:
Are you still seeing bugs with the latest versions of react native without disabling view recycling? If yes, I will merge this. 

Is there a reproducible repo that I can look for the issue?

fadookie:
@matinzd Sorry, I didn't try to isolate a repro case. My company's security team has temporarily blocked engineers from pushing code to GitHub while they figure out some sort of new policy to allow that so I'd be blocked on that anyway for the time being. I can report back on this with our app the next time we upgrade React Native and see what happens if we remove this patch, but we have not upgraded since I opened this issue.

Issue #1066, opened 2023-07-26
Crash when provided lottie link is broken


arasrezaei:
Hi there,
i've got this error after the link of lottie animation is broken or networking is not available:
```
Your app just crashed. See the error below.
java.lang.IllegalStateException: Unable to parse composition
  com.airbnb.lottie.LottieAnimationView.lambda$static$0(LottieAnimationView.java:75)
  com.airbnb.lottie.LottieAnimationView$$ExternalSyntheticLambda1.onResult(Unknown Source:2)
  com.airbnb.lottie.LottieAnimationView$WeakFailureListener.onResult(LottieAnimationView.java:117)
  com.airbnb.lottie.LottieAnimationView$WeakFailureListener.onResult(LottieAnimationView.java:99)
  com.airbnb.lottie.LottieTask.notifyFailureListeners(LottieTask.java:162)
  com.airbnb.lottie.LottieTask.lambda$notifyListeners$0$com-airbnb-lottie-LottieTask(LottieTask.java:138)        
  com.airbnb.lottie.LottieTask$$ExternalSyntheticLambda0.run(Unknown Source:2)
  android.os.Handler.handleCallback(Handler.java:873)
  android.os.Handler.dispatchMessage(Handler.java:99)
  android.os.Looper.loop(Looper.java:193)
  android.app.ActivityThread.main(ActivityThread.java:6669)
  java.lang.reflect.Method.invoke(Native Method)
  com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
  com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
Caused by java.io.FileNotFoundException: https://[**brokenlink**]/categories/a4588f9e-b5fa-4dd6-aea7-11730889b6b4.json
  android.content.res.AssetManager.nativeOpenAsset(Native Method)
  android.content.res.AssetManager.open(AssetManager.java:744)
  android.content.res.AssetManager.open(AssetManager.java:721)
  com.airbnb.lottie.LottieCompositionFactory.fromAssetSync(LottieCompositionFactory.java:220)
  com.airbnb.lottie.LottieCompositionFactory.lambda$fromAsset$1(LottieCompositionFactory.java:187)
  com.airbnb.lottie.LottieCompositionFactory$$ExternalSyntheticLambda7.call(Unknown Source:6)
  java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #1416, opened 2026-04-02
fix: add file.exists() check for file:// URIs to fix .lottie crashes on CodePush

adityasid:
## Description

Adds a missing `file.exists()` guard before opening a `FileInputStream` in the `file://` URI handler of `LottieAnimationViewPropertyManager.kt`.

Without this check, a `FileNotFoundException` propagates uncaught through Fabric's view pre-allocation pipeline, causing a fatal crash on the main thread. This affects apps using OTA update tools (CodePush, Expo Updates) where `.lottie` assets are loaded from the local filesystem via `file://` URIs.

## Problem

The `file://` URI handler at lines 150–157 opens a `FileInputStream` directly without verifying the file exists:

```kotlin
if (scheme == "file") {
    val uri = Uri.parse(assetName)
    uri.path?.let { path ->
        val fileWithScheme = File(path)
        view.setAnimation(
            ZipInputStream(FileInputStream(fileWithScheme)), // crashes if file is unavailable
            assetName.hashCode().toString()
        )
    }
}
```

The direct file path handler just above (lines 136–144) already guards against this correctly:

```kotlin
var file = File(assetName)
if (file.exists()) { // ✅ guarded
    view.setAnimation(ZipInputStream(FileInputStream(file)), ...)
}
```

### Key observations

- The crash does **not** happen for all users. The same CodePush bundle works for the majority of users. Only a subset of users (likely those with slower I/O, low storage, or specific OEMs) hit this crash.
- CodePush hash verification confirms **all files are present at install time**, but the file can become transiently unavailable when the `LottieView` is pre-allocated during Fabric's frame callback on the next app restart.
- Other React Native native components (e.g., `Image` via Fresco) handle `FileNotFoundException` gracefully. Only `LottieView` crashes fatally because of the missing guard.

## Fix

Added a `fileWithScheme.exists()` check before `FileInputStream`, matching the existing pattern. If the file is unavailable, the animation is skipped with a log message instead of crashing.

## Crash log (before fix)

```
FATAL EXCEPTION: main
java.io.FileNotFoundException:
  /data/user/0/.../files/CodePush/.../CodePush/raw/src_assets_lottie_loader.lottie:
  open failed: ENOENT (No such file or directory)
  at com.airbnb.android.react.lottie.LottieAnimationViewPropertyManager.commitChanges(LottieAnimationViewPropertyManager.kt:155)
  at com.airbnb.android.react.lottie.LottieAnimationViewManagerImpl.setSourceDotLottieURI(LottieAnimationViewManagerImpl.kt:205)
  at com.facebook.react.fabric.mounting.mountitems.PreAllocateViewMountItem.execute(PreAllocateViewMountItem.kt:39)
  at com.facebook.react.fabric.FabricUIManager$DispatchUIFrameCallback.doFrameGuarded(FabricUIManager.java:1513)
```

## Related

- Fixes the crash scenario described in #1098
- The `file://` handler was introduced in #1319
- Similar crash pattern as #1066


matinzd (approved):
Thanks!

adityasid:
@matinzd could you please merge it?

adityasid:
@matinzd could you please review and merge it?

PR #1419, opened 2026-04-09
fix(android): parse dotLottie files as ZIP instead of relying on auto-detection

### LINKED ISSUE #1394: Android: dotLottie (.lottie) files fail to load with New Architecture - parsed as JSON instead of ZIP
## Description

When using `.lottie` files (dotLottie format) with React Native's New Architecture (Fabric) on Android, the animations fail to load with the error:

```
Use JsonReader.setLenient(true) to accept malformed JSON at path $
```

The native code is attempting to parse the dotLottie file as JSON instead of recognizing it as a ZIP archive.

## Steps to Reproduce

1. Create a React Native app with New Architecture enabled (`newArchEnabled=true`)
2. Add a `.lottie` file to assets
3. Use it with LottieView:
```tsx
import Animation from './animation.lottie';

<LottieView source={Animation} autoPlay loop />
```

## Expected Behavior

The dotLottie file should be detected as a ZIP archive and extracted properly.

## Actual Behavior

The file is passed to JSON parser, which fails because ZIP files start with `PK` bytes, not valid JSON.

## Root Cause Analysis

The bug is in the Android native code (`LottieAnimationViewPropertyManager.kt` lines 146-161):

```kotlin
val scheme = runCatching { Uri.parse(assetName).scheme }.getOrNull()
if (scheme != null) {
    if (scheme == "file") {
        // Uses ZipInputStream - CORRECT ✓
        view.setAnimation(ZipInputStream(FileInputStream(fileWithScheme)), ...)
    } else {
        // For http://, asset://, etc.
        view.setAnimationFromUrl(assetName)  // BUG: doesn't handle dotLottie!
    }
}
```

When a `.lottie` asset is loaded via `require()`, the resolved URI has a scheme like `http://localhost:8081/...` (dev mode) or `asset://...` (release). Since this isn't `file://`, the code calls `view.setAnimationFromUrl()`.

The underlying Lottie Android library's `setAnimationFromUrl()` doesn't detect dotLottie format - it assumes JSON and tries to parse directly, causing the "malformed JSON" error.

## Suggested Fix

Before calling `setAnimationFromUrl()`, check if the URL ends with `.lottie` and handle it appropriately:

```kotlin
} else {
    if (assetName.endsWith(".lottie")) {
        // Fetch URL content and wrap in ZipInputStream
        // Or use a dotLottie-aware loading method
    } else {
        view.setAnimationFromUrl(assetName)
    }
}
```

Alternatively, fetch the content first and check for ZIP magic bytes (`PK` / `0x504B`) before deciding how to parse.

## Workaround

Extract the JSON from `.lottie` files and use `.json` files instead:
```bash
unzip -p animation.lottie "animations/*.json" > animation.json
```

## Environment

- lottie-react-native: 7.3.4
- React Native: 0.81.5
- New Architecture: enabled
- Platform: Android

edkimmel:
## Summary

Fixes #1394

When a `.lottie` file is loaded via `sourceDotLottieURI` with a non-file URI scheme (http/https from Metro dev server, CDN URLs, etc.), `setAnimationFromUrl()` delegates to Lottie Android's `NetworkFetcher`, which determines format via **Content-Type header and URL extension** — not magic byte detection. When the server doesn't set `application/zip` Content-Type (common with Metro dev server, CDNs serving `application/octet-stream`), the content is incorrectly parsed as JSON:

```
Use JsonReader.setLenient(true) to accept malformed JSON at path $
```

Since `sourceDotLottieURI` already guarantees the content is dotLottie (ZIP) format, this commit:

- **URL path**: Replaces `setAnimationFromUrl()` with an explicit background download + `LottieCompositionFactory.fromZipStreamSync()` that always parses as ZIP
- **Raw resource path**: Replaces `setAnimation(resourceId)` with `ZipInputStream(openRawResource())` — the `fromRawRes()` magic byte detection can fail if AAPT2 reprocesses the resource

Both the `file://` and local file paths already used explicit `ZipInputStream` wrapping and were unaffected.

### Root cause

Decompiling `NetworkFetcher.fromInputStream` in lottie-android 6.7.1 confirms the detection logic:

```java
// Defaults to "application/json" when Content-Type is null
if (contentType == null) contentType = "application/json";

// Only treats as ZIP if Content-Type is application/zip OR URL ends with .lottie
if (contentType.contains("application/zip") || 
    url.split("\\?")[0].endsWith(".lottie")) {
    // handle as zip
} else {
    // falls through to JSON parsing ← BUG for dotLottie files
}
```

This differs from `LottieCompositionFactory.fromRawResSync` and `fromInputStreamSync`, which use magic byte detection (`PK` / `0x504B0304`). The `sourceDotLottieURI` URL path goes through `NetworkFetcher`, not `fromInputStreamSync`.

## Test plan

- [x] Load a `.lottie` file via `require('./animation.lottie')` on Android (debug mode — Metro serves over HTTP)
- [x] Load a `.lottie` file via remote URL on Android
- [x] Verify `.lottie` files still load correctly on iOS (unchanged)
- [x] Verify `.json` Lottie files still load correctly on Android via `sourceURL` path (unchanged)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


edkimmel:
We were bit by this in production, sad to see that the original issue was closed as stale, when it was a real and valid problem just not acknowledged by the lottie team.

matinzd:
> We were bit by this in production, sad to see that the original issue was closed as stale, when it was a real and valid problem just not acknowledged by the lottie team.

We are not part of the Airbnb-maintained Lottie team. This repository serves as the React Native SDK. Unfortunately, I did not have time to review these issues earlier, and our stale bot was misconfigured, which caused them to be closed automatically. I have reopened the issues and will also take a look at this PR.

PR #1438, opened 2026-07-12
feat: expose Android opacity layer option

### LINKED ISSUE #1436: Android applies precomp/group opacity per-layer — overlapping layers bleed through; expose applyingOpacityToLayersEnabled
### Summary

On Android, a **precomp/group opacity** over **overlapping child layers** is applied to each child layer independently rather than to the composited group. Whenever the group is drawn at reduced opacity, the overlapping children show through one another. iOS and web composite the group first, so they render correctly. It's most noticeable during **cross-dissolves** (animated group
opacity), but it also reproduces with a plain **static** reduced opacity — which the minimal repro below uses so the effect is deterministic and easy to inspect.

`lottie-android` already has the fix — `LottieAnimationView.setApplyingOpacityToLayersEnabled(true)` — but `lottie-react-native` doesn't expose it as a prop, so there's no way to opt in from JS.

### Environment

| | |
|---|---|
| `lottie-react-native` | 7.3.8 (also reproduced on 7.3.1 via the Snack) |
| `lottie-android` (bundled) | 6.7.1 |
| `react-native` | 0.81.5 (New Architecture / Fabric) |
| Platforms affected | Android only (iOS + web render correctly) |

The bug is version-independent — it stems from a `lottie-android` default (opacity applied per shape, off by default) present across its 6.x/7.x line, so any recent `lottie-react-native` exhibits it.

### Expected

The group should be composited first, then have its opacity applied to the result as a whole — so overlapping child layers never show through one another. In a cross-dissolve the outgoing and incoming layers fade as a single unit, with no ghosting or double-exposure mid-transition, exactly as iOS and web already render it.

### Actual (Android)

**Crossfade in progress, showing bleedthrough bug.**
Crossfade in progress showing hair layer bleedthrough

**After the crossfade completes — resolved:**

Crossfade finished showing hair layer resolved

### Root cause

`lottie-android` applies opacity to shapes by default. For overlapping translucent content it must instead composite the layers off-screen and then apply opacity to the result. That mode exists on `LottieDrawable` (and `LottieAnimationView` delegates to it):

```java
// com.airbnb.lottie.LottieDrawable
/**
 * Sets whether to apply opacity to the each layer instead of shape.
 * <p>
 * Opacity is normally applied directly to a shape. In cases where translucent shapes overlap, applying opacity to a layer will be more accurate
 * at the expense of performance.
 * <p>
 * The default value is false.
 * <p>
 * Note: This process is very expensive. The performance impact will be reduced when hardware acceleration is enabled.
 *
 * @see android.view.View#setLayerType(int, android.graphics.Paint)
 * @see LottieAnimationView#setRenderMode(RenderMode)
 */
public void setApplyingOpacityToLayersEnabled(boolean isApplyingOpacityToLayersEnabled)
```

`lottie-react-native` never calls it and exposes no prop for it (only `renderMode`, which doesn't address this).

### Proposed fix

Expose a prop — e.g. `applyOpacityToLayersAndroid?: boolean` (default `false` to preserve current performance) — wired through the view manager to `setApplyingOpacityToLayersEnabled(...)`, mirroring how `enableMergePathsAndroidForKitKatAndAbove` is plumbed. iOS already composites correctly, so the prop can be a no-op there (or documented as Android-only).

### Workaround

Until then we patch the Android view manager to enable it at view creation (`patch-package`):

```kotlin
// LottieAnimationViewManagerImpl.createViewInstance
LottieAnimationView(context).apply {
    scaleType = ImageView.ScaleType.CENTER_INSIDE
    setApplyingOpacityToLayersEnabled(true)
}
```

### Steps to reproduce

1. Open the minimal reproduction — clone the linked gist, or create a new Expo Snack (SDK 54) with its two files (`App.js` + `bleedthrough.json`, both at the project root as siblings).
2. Add the dependency `lottie-react-native@~7.3.1` (SDK 54's recommended version). Do NOT add `@lottiefiles/dotlottie-react` — it's the optional web-only peer, not needed for Android; the peer warning is safe to ignore.
3. `App.js` renders the animation:
   `<LottieView source={require('./bleedthrough.json')} autoPlay loop style={{ width: 300, height: 300 }} />`
4. `bleedthrough.json` is a precomp (`overlap`) containing two overlapping opaque 200×200 rectangles — **blue** on top, **red** beneath, offset so they overlap in the middle. The precomp layer's opacity is a **static 40%** (`ks.o = 40`). No animation is involved.
5. Open the Snack on a **physical Android device or emulator via Expo Go** — use the Android / "My Device" tab, NOT the Web preview.
6. Observe the rendered graphic: three vertical bands — **pink** (red-only, left), a **middle overlap band**, and **light blue** (blue-only, right).

**Result:** the middle overlap band renders **purple** (the red rectangle showing through the blue) instead of the light blue you'd get if the group were composited before opacity was applied. On iOS/web — or on Android with `setApplyingOpacityToLayersEnabled(true)` — that band is light blue.

https://gist.github.com/SchramboVida/4f41cd7873802f13eb6d5d49e38100d9

7.3.8

### React Native version

0.81.5

### Platforms

Android, iOS, Web

### Workflow

React Native

### Architecture

Fabric (New Architecture)

### Build type

Release app & production bundle

### Device

Real device

### Acknowledgements

Yes

wanxiankai:
## Summary

Closes #1436.

This exposes Android's `LottieAnimationView.setApplyingOpacityToLayersEnabled(...)` as a React Native prop:

```tsx
<LottieView applyOpacityToLayersAndroid />
```

The new prop is named `applyOpacityToLayersAndroid` and defaults to `false` to preserve the current rendering/performance behavior. When enabled, Android applies opacity to composited layers instead of individual shapes, which avoids bleed-through artifacts for overlapping translucent/precomp layers.

## What changed

- Added `applyOpacityToLayersAndroid?: boolean` to the public TypeScript props.
- Added the prop to the native component codegen spec.
- Wired the prop through both Android old architecture and new architecture managers.
- Applied it in `LottieAnimationViewPropertyManager` via `setApplyingOpacityToLayersEnabled`.

## Test plan

Passed locally:

- `yarn install --immutable`
- `yarn workspace lottie-react-native build`
- `yarn tsc`
- `git diff --check`

Attempted but blocked by local environment:

- `yarn workspace lottie-react-native exec eslint src --ext .ts,.tsx`
  - Failed because the existing ESLint config references missing dependency `@react-native/eslint-plugin-specs`.
- `./gradlew compileDebugKotlin`
  - Gradle dependency resolution/configuration did not complete reliably in this sandbox. Offline mode also failed because `com.android.tools.build:gradle:7.2.2` and `org.jetbrains.kotlin:kotlin-gradle-plugin:1.7.10` were not cached.

## Compatibility

- Android: implemented
- iOS: not changed
- Web: not changed
- Windows: not changed

### DIFF
--- a/packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManagerImpl.kt
+++ b/packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewManagerImpl.kt
@@ -307,6 +307,14 @@ internal object LottieAnimationViewManagerImpl {
     }
 
     @JvmStatic
+    fun setApplyOpacityToLayers(
+        applyOpacityToLayers: Boolean,
+        viewManager: LottieAnimationViewPropertyManager
+    ) {
+        viewManager.applyOpacityToLayers = applyOpacityToLayers
+    }
+
+    @JvmStatic
     fun setEnableSafeMode(
         enableSafeMode: Boolean,
         viewManager: LottieAnimationViewPropertyManager
--- a/packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
+++ b/packages/core/android/src/main/java/com/airbnb/android/react/lottie/LottieAnimationViewPropertyManager.kt
@@ -55,6 +55,7 @@ class LottieAnimationViewPropertyManager(view: LottieAnimationView) {
     var scaleType: ImageView.ScaleType? = null
     var imageAssetsFolder: String? = null
     var enableMergePaths: Boolean? = null
+    var applyOpacityToLayers: Boolean? = null
     var enableSafeMode: Boolean? = null
     var colorFilters: ReadableArray? = null
     var textFilters: ReadableArray? = null
@@ -226,6 +227,11 @@ class LottieAnimationViewPropertyManager(view: LottieAnimationView) {
         enableMergePaths?.let {
             view.enableMergePathsForKitKatAndAbove(it)
             enableMergePaths = null
+        }
+
+        applyOpacityToLayers?.let {
+            view.setApplyingOpacityToLayersEnabled(it)
+            applyOpacityToLayers = null
         }
 
         enableSafeMode?.let {
--- a/packages/core/android/src/newarch/com/airbnb/android/react/lottie/LottieAnimationViewManager.kt
+++ b/packages/core/android/src/newarch/com/airbnb/android/react/lottie/LottieAnimationViewManager.kt
@@ -1,6 +1,7 @@
 package com.airbnb.android.react.lottie
 
 import android.animation.Animator
+import com.airbnb.android.react.lottie.LottieAnimationViewManagerImpl.setApplyOpacityToLayers
 import com.airbnb.android.react.lottie.LottieAnimationViewManagerImpl.setColorFilters
 import com.airbnb.android.react.lottie.LottieAnimationViewManagerImpl.setEnableMergePaths
 import com.airbnb.android.react.lottie.LottieAnimationViewManagerImpl.setEnableSafeMode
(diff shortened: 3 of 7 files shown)


matinzd (approved):

matinzd:
Is this ready for go?

@wanxiankai

wanxiankai:
Yes — this is ready to go. I've marked the PR as ready for review. The remaining required workflows are waiting for maintainer approval to run; the branch otherwise merges cleanly. Thanks!

PR #1452, opened 2026-08-05
chore(deps): upgrade react-native-test-app to 5.4.7

TheRogue76:
Bumps `react-native-test-app` from 4.1.4 to 5.4.7 in the example app, and
drops the stale 3.8.7 devDependency from the monorepo root. Nothing outside
`example/` referenced it, and keeping a second major version around meant
yarn installed two copies of the package.

The only breaking change in 5.0.0 is dropping support for React Native 0.75
and older; the example is on 0.78.3, so no migration was needed. `app.json`
still validates against the 5.4.7 schema and `configureProjects` is
unchanged, so no config updates were required.

5.2.0 raised the minimum Node version to 20.19, so `engines.node` and the
CI workflows move off Node 18. CI goes to 22 (active LTS) rather than the
bare 20.19.4 floor.

Verified with `yarn install`, `yarn setup`, `yarn tsc`, and `pod install`
for the example app (70/70 pods).

---

### DIFF
--- a/.github/workflows/android-fabric-build.yml
+++ b/.github/workflows/android-fabric-build.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 22
 
       - name: Enable corepack
         run: corepack enable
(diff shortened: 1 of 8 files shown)


TheRogue76:
Superseded by #1453.

The `react-native-test-app` upgrade can't land on its own: 5.4.7 pins `androidx.camera` to 1.6.1, which needs compileSdk 36 + AGP 8.9.1+, and RN 0.78 supplies compileSdk 35 / AGP 8.8.0. RNTA 4.1.4 also caps `react-native` at 0.78, so neither version moves independently — they're combined into a single change in #1453, along with raising the minimum React Native version to 0.84.

PR #1456, opened 2026-08-05
feat: scaffold v8 package and example-v8 app

TheRogue76:
Structure only, deliberately not wired to anything. v7, `example/` and CI are
untouched; nothing imports the new package except the new app.

`packages/nitro` (`lottie-react-native-nitro`, private) declares a Nitro
`HybridView` named `LottieView` with a single placeholder prop, and Swift and
Kotlin stubs that compile and render an empty native view. No Lottie rendering
-- that is the next change. Everything except the package's `name` is
final-shaped: the view name, spec path, C++/Swift/Kotlin namespaces and class
names all match where v8 is going. The name has to differ for now because yarn
throws on duplicate workspace names and `packages/core` already claims
`lottie-react-native`.

`private: true` matters: the root `@release-it-plugins/workspaces` config globs
`packages/*`, so without it an incomplete package would enter the release flow.

`example-v8` is a hand-authored react-native-test-app app (ios, android, web)
pinned to the same versions as `example/`. It was authored rather than generated
with `create-react-native-library` because the generator brings its own React
Native version, CI and lint config, which would fight the 0.84.1 / RNTA 5.4.7 /
Gradle 9.0 setup.

Four things that are load-bearing and not obvious:

- `LottieNitroPackage.kt` is required three times over. Nitro's autolinking
  generates constructor plumbing but does not register views on Android. The
  class must be named `*Package` or the Community CLI never autolinks the
  library at all; it must return the generated view manager; and its companion
  initialiser must force `System.loadLibrary`, because the Fabric
  ComponentDescriptor is only registered at `JNI_OnLoad`. Without that last
  part React Native silently falls back to
  `UnimplementedNativeViewComponentDescriptor`.
- The Android `build.gradle` must not declare `libraryName` or apply
  `com.facebook.react`. React Native's autolinker emits `add_subdirectory(...)`
  for any dependency declaring a `libraryName`, pointing CMake at a codegen
  directory that is never generated here -- this module uses nitrogen, not RN
  codegen. Verified: `react-native config` reports `libraryName: null` for it.
- `newArchEnabled=true` lives in `example-v8/android/gradle.properties` rather
  than the invocation, because `react-native-nitro-modules` gates its setup on
  the *root* project property.
- `src/index.web.tsx` is required, not cosmetic. `getHostComponent` deep-imports
  `react-native/Libraries/NativeComponent/NativeComponentRegistry`, which is
  Flow source webpack cannot transform. Confirmed the shim is what resolves and
  that the web bundle contains no nitro-modules code.

`nitrogen/generated/` is committed because the podspec `load`s a generated .rb,
the gradle file `apply from:`s a generated .gradle, CMakeLists `include`s a
generated .cmake and the view wrapper imports a generated .json -- a fresh clone
cannot even `pod install` without them.

Verified: nitrogen generates 1/1 HybridObject; both typechecks pass; Android
library and app both `BUILD SUCCESSFUL` with `libLottieNitro.so` and
`libNitroModules.so` packaged in the APK; iOS `BUILD SUCCEEDED` with the Swift
stub compiled into `libLottieNitro.a`; the web bundle compiles. Nitro 0.36.5
turned out to build fine on Gradle 9.0 / AGP 8.8.2 despite targeting 9.3.1, so
no AGP pin was needed. `yarn setup` and `yarn tsc` still pass for v7.

No CI workflows are added.

---


packages/nitro/nitrogen/generated/ios/LottieNitro+autolinking.rb
matinzd, line 1: What's the best practice for generated files from nitro? Shouldn't we ignore them in .gitignore and generate them on the fly?

For e.g this library has nitrogen folder in gitignore:

- https://github.com/ChristopherGabba/react-native-nitro-audio-manager/blob/a89a43f8940db71dd3e5dc7759ae73fb03f7d17c/.gitignore#L88

matinzd (commented):
Other than the generated files, which I think are fine given that most Nitro libraries include generated files in the repo, I’m good with the changes. 🚀

packages/nitro/nitrogen/generated/ios/LottieNitro+autolinking.rb
TheRogue76: Couple of things that make me doubt that is a good idea:

1. It's not that much code, so there isn't any gains from not including it
2. Realistically, anyone pulling the repo would like to just get working implementing fixes and changes, not setting up the code generation, and at least at the moment i don't see a reason not to include it. It also helps prevent drifts if Nitro's code generation ever changes, so we can track it

matinzd (approved):

PR #1458, opened 2026-08-05
feat: v8 props impl

TheRogue76:


example-v8/ios/Podfile
```diff
@@ -6,8 +6,39 @@ require "#{ws_dir}/node_modules/react-native-test-app/test_app.rb"
...
+      bundle_target.build_configurations.each do |config|
+        current = config.build_settings['IPHONEOS_DEPLOYMENT_TARGET']
+        next if current && current.to_f >= minimum.to_f
+
+        config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = minimum
+      end
+    end
+  end
+end
+
```
matinzd: I think the example should work without this patch. Can you try again without this?

packages/nitro/android/src/main/java/com/margelo/nitro/lottienitro/LottieColorParser.kt
```diff
@@ -0,0 +1,106 @@
...
+    val b = channel(parts[2]) ?: return null
+    val a = if (parts.size == 4) alpha(parts[3]) ?: return null else 255
+    return Color.argb(a, r, g, b)
+  }
+
+  /**
+   * CSS Level 1 names. Kept deliberately small and identical to the Swift side
+   * rather than pulling in the full CSS list, so the two can be diffed by eye.
+   */
+  private val NAMED_COLORS: Map<String, Int> = mapOf(
```
matinzd: I think react native has a helper for this. I will check and get back.

example-v8/ios/Podfile
TheRogue76: It gave a bunch of weird errors last time around. I had to do this for V7 and V8. I'll try again and if i see it, i will post the error again in this thread. It was complaining about lottie iOS's version then

TheRogue76: Nope, it fails, it is the build that fails, cause Lottie iOS's privacy info seems to target iOS 13:
```
The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 13.0,                                                                                                                        
  but the range of supported deployment target versions is 15.0 to 27.0.x.                                                                                                                                       
  (in target 'lottie-ios-LottiePrivacyInfo' from project 'Pods')
```

TheRogue76: It should go away once we are a SPM only package but for now, not a lot of ideas on how to override it

packages/nitro/android/src/main/java/com/margelo/nitro/lottienitro/LottieColorParser.kt
TheRogue76: Yeah found it again, big ass goof. Fixed under https://github.com/lottie-react-native/lottie-react-native/pull/1458/commits/643ffac8c700c8cc8acfee4252270fcbb835a6ff

TheRogue76:
Got rid of the slop comments, was way to excited when i started working on this to establish ground rules with the clanker not to write so much crap. Left an Agent file with specific instructions that it can only yap in the Architecture document

matinzd (approved):

example-v8/ios/Podfile
matinzd: Probably it's due to xcode 27. We can ignore it.

PR #1463, opened 2026-08-13
fix(android): skip explicit kotlin-android plugin when AGP 9 built-in Kotlin is enabled

kimchi-developer:
## What

`packages/core/android/build.gradle` applies `kotlin-android` unconditionally. This makes the
library fail to configure on Android Gradle Plugin 9, where AGP compiles Kotlin itself. This PR
applies the plugin only when AGP's built-in Kotlin support is *not* in effect.

## The bug

AGP 9.0 ships built-in Kotlin support and [enables it by default for every module that applies
AGP](https://developer.android.com/build/migrate-to-built-in-kotlin) — no `org.jetbrains.kotlin.android`
/ `kotlin-android` plugin needed. A module that still applies it explicitly now fails at
configuration time with one of:

```
Failed to apply plugin 'org.jetbrains.kotlin.android'.
    > Cannot add extension with name 'kotlin', as there is an extension already registered with that name.
```

```
Failed to apply plugin 'org.jetbrains.kotlin.android'
    > The 'org.jetbrains.kotlin.android' plugin is no longer required for Kotlin support since AGP 9.0.
```

Which of the two you get depends on the AGP/KGP pair, but either way the build stops while
configuring `:lottie-react-native`. Because the failure comes from a *dependency's* build file, the
only fix available to an app developer today is the global escape hatch
`android.builtInKotlin=false` in their `gradle.properties` — which also forces `android.newDsl=false`
and, per the AGP docs, stops working in AGP 10.0.

The repo's own example apps are not affected because they already set `android.builtInKotlin=false`
(`example/android/gradle.properties`, added for `react-native-test-app`), so CI does not currently
exercise the failing path.

## The fix

```groovy
def isBuiltInKotlinEnabled() {
    def agpMajorVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger()
    if (agpMajorVersion < 9) {
        return false
    }
    def builtInKotlinProperty = project.findProperty('android.builtInKotlin')
    return builtInKotlinProperty == null || builtInKotlinProperty.toString().toBoolean()
}

apply plugin: 'com.android.library'

if (!isBuiltInKotlinEnabled()) {
    apply plugin: 'kotlin-android'
}
```

Why the condition has both halves:

- **AGP major `>= 9`** — built-in Kotlin only exists from AGP 9 onwards. Below that the explicit
  plugin is still mandatory, so the guard must never skip it there.
- **`android.builtInKotlin` unset or `true`** — the flag is the documented opt-out. A consumer that
  sets `android.builtInKotlin=false` (as this repo's own examples do) is back to needing the
  explicit plugin even on AGP 9. Absent means "on", since AGP 9 defaults it to enabled.

Simply *removing* `apply plugin: 'kotlin-android'` would not work: it breaks every AGP 8 consumer
and also the AGP 9 consumers who opted out.

## Compatibility

No behaviour change on any previously supported configuration:

| AGP | `android.builtInKotlin` | `kotlin-android` applied |
|---|---|---|
| 7.x / 8.x | unset | yes (unchanged) |
| 7.x / 8.x | `false` | yes (unchanged) |
| 9.x | unset | **no** (AGP compiles Kotlin) |
| 9.x | `false` | yes |
| 9.x | `true` | **no** |

`com.android.Version` has been on the AGP classpath since AGP 7.0, and this file's own `buildscript`
block already pins `com.android.tools.build:gradle:7.2.2`, so the reference resolves on every AGP
version this library can be built with. Version strings with a qualifier (`9.0.0-alpha01`) parse
correctly because only the first `.`-separated token is read.

The `kotlin-gradle-plugin` classpath entry in `buildscript` is left in place — having it on the
classpath without applying the plugin is harmless, and it is still needed for the non-built-in
cases.

## Testing

The guard's truth table was verified by running the same Groovy in a Gradle 9.0.0 build script with
the AGP version stubbed, across `7.2.2` / `8.13.0` / `9.0.0` / `9.0.0-alpha01` / `10.1.2` and with
`android.builtInKotlin` unset / `false` / `true`; results match the table above.

## Out of scope

`packages/nitro/android/build.gradle` applies `org.jetbrains.kotlin.android` the same way and has
the same problem, but that package is `private: true` and not published yet, so I left it alone to
keep this fix reviewable. Happy to include it in this PR if you'd prefer them changed together.

### DIFF
--- a/packages/core/android/build.gradle
+++ b/packages/core/android/build.gradle
@@ -25,8 +25,24 @@ def isNewArchitectureEnabled() {
     return project.hasProperty("newArchEnabled") && project.newArchEnabled == "true"
 }
 
+// AGP 9.0 compiles Kotlin itself and enables that built-in support by default,
+// so applying the standalone Kotlin Android plugin on top of it fails the build.
+// Applications can still opt out with `android.builtInKotlin=false`, in which
+// case the explicit plugin is required again, as it is on every AGP below 9.
+def isBuiltInKotlinEnabled() {
+    def agpMajorVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger()
+    if (agpMajorVersion < 9) {
+        return false
+    }
+    def builtInKotlinProperty = project.findProperty('android.builtInKotlin')
+    return builtInKotlinProperty == null || builtInKotlinProperty.toString().toBoolean()
+}
+
 apply plugin: 'com.android.library'
-apply plugin: 'kotlin-android'
+
+if (!isBuiltInKotlinEnabled()) {
+    apply plugin: 'kotlin-android'
+}
 
 if (isNewArchitectureEnabled()) {
     apply plugin: 'com.facebook.react'
--- a/packages/core/android/build.gradle
+++ b/packages/core/android/build.gradle
@@ -25,10 +25,6 @@ def isNewArchitectureEnabled() {
     return project.hasProperty("newArchEnabled") && project.newArchEnabled == "true"
 }
 
-// AGP 9.0 compiles Kotlin itself and enables that built-in support by default,
-// so applying the standalone Kotlin Android plugin on top of it fails the build.
-// Applications can still opt out with `android.builtInKotlin=false`, in which
-// case the explicit plugin is required again, as it is on every AGP below 9.
 def isBuiltInKotlinEnabled() {
     def agpMajorVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger()
     if (agpMajorVersion < 9) {
--- a/packages/nitro/android/build.gradle
+++ b/packages/nitro/android/build.gradle
@@ -32,8 +32,20 @@ def getExtOrIntegerDefault(name) {
         : (project.properties["LottieNitro_" + name]).toInteger()
 }
 
+def isBuiltInKotlinEnabled() {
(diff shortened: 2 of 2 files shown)


packages/core/android/build.gradle
```diff
@@ -25,8 +25,24 @@ def isNewArchitectureEnabled() {
     return project.hasProperty("newArchEnabled") && project.newArchEnabled == "true"
 }
 
+// AGP 9.0 compiles Kotlin itself and enables that built-in support by default,
+// so applying the standalone Kotlin Android plugin on top of it fails the build.
+// Applications can still opt out with `android.builtInKotlin=false`, in which
+// case the explicit plugin is required again, as it is on every AGP below 9.
```
TheRogue76: The comment is not super needed. Please delete it

TheRogue76:
Hi @kimchi-developer . Thank you for the contribution, please do the same for the nitro package as well.

packages/core/android/build.gradle
kimchi-developer: Removed, thanks.

kimchi-developer:
Thanks for the review @TheRogue76! Both points addressed:

- Dropped the comment in `packages/core/android/build.gradle`.
- Applied the same guard to `packages/nitro/android/build.gradle` — it applies the plugin under its full id (`org.jetbrains.kotlin.android`), but that is the same plugin and hits the identical AGP 9 failure. Those two are the only places in the repo that apply it.

Also rebased onto `master`, so the branch is no longer behind.

TheRogue76 (approved):

PR #1465, opened 2026-08-17
switch to lottie spm for our SPM support

TheRogue76:
Yeah i goofed and forgot that lottie has a separate repo for their spm which is muuuuuch lighter


matinzd:
Can we add CI for checking SPM integration?

TheRogue76:
> Can we add CI for checking SPM integration?

I did think about it, problem was that the best way i could think for it to be reliable was adding a new example, and between example and example-v8 it is getting crowded, i want a better system for it, but i am open to ideas about how we can handle this

matinzd (approved):

